PR #16, opened 2020-05-14
Add support for virtual steps that accumulate per-sample or clipped gradients

ftramer:
## Types of changes

The current codebase can only handle batch sizes that fit entirely in memory. However, DP-SGD might benefit from training with larger batch sizes than for standard models (e.g., [the best MNIST results in tensorflow/privacy use a batch size of 256](https://github.com/tensorflow/privacy/tree/master/tutorials#select-parameters))

For training without DP, an easy way to support large batch-sizes is roughly as follows:
```python
for i, (x, y) in enumerate(train_loader):  # samples batches of size `B`
    y_hat = model(x)
    loss = criterion(y_hat, y)
    loss.backward()

    if (i+1) % N == 0:  # only takes a step after processing `N*B` examples
        optimizer.step()
        optimizer.zero_grad()
```
By repeatedly calling `loss.backward()` for many mini-batches, the gradient contributions of each mini-batch are summed-up in each weight's `.grad` field.
This PR implements two complementary approaches to extend this idea to DP-SGD:

1) Calling `loss.backward()` repeatedly now stores the per-sample gradients for all mini-batches (previously, a new call to `loss.backward()` would overwrite any existing per-sample gradients).
If we call `loss.backward()` `N` times on mini-batches of size `B`, then each weight's `.grad_sample` field will contain `N*B` gradients.
Then, when calling `optimizer.step()`, the privacy-engine clips all `N*B` gradients and computes the average gradient for an effective batch of size `N*B`. A call to `optimizer.zero_grad()` erases the per-sample gradients.

2) By calling `optimizer.virtual_step()` after `loss.backward()`, the `B` per-sample gradients for this mini-batch are clipped and summed up into a gradient accumulator. The per-sample gradients can then be discarded. After `N` iterations (alternating calls to `loss.backward()` and `optimizer.virtual_step()`), a call to `optimizer.step()` will compute the average gradient for an effective batch of size `N*B`

The main benefit of option 2) is that it is much more memory efficient: it discards the per-sample gradients after every mini-batch. It can thus handle batches of arbitrary size. The downside is that per-sample gradients have to be clipped before the entire effective batch has been processed, which might limit some dynamic clipping strategies.

The code for option 2) looks like this:

```python
privacy_engine = PrivacyEngine(
    model,
    batch_size=B * N,  # effective batch size
    ...
)
privacy_engine.attach(optimizer)

for i, (x, y) in enumerate(train_loader):  # samples batches of size `B`
    y_hat = model(x)
    loss = criterion(y_hat, y)
    loss.backward()
    optimizer.virtual_step()  # clip and accumulate `B` gradients

    if (i+1) % N == 0:  # only takes a step after processing `N*B` examples
        optimizer.step()
        optimizer.zero_grad()
```

The test cases in `test/virtual_step_test.py` cover the new functionality.
Both options 1) and 2) above have been tested for `examples/imagenet.py`. There is a new command-line argument, `effective_batch_size`, which controls the size of batches consumed for each optimizer step (i.e., `N*B` above).

Relying on option 1) did not help much in my experiments, because storing all the per-sample gradients remains very memory intensive. With option 2) however, it is possible to train with effective batch sizes above 1000.

## Checklist

### DIFF
--- a/examples/imagenet.py
+++ b/examples/imagenet.py
@@ -80,14 +80,12 @@ parser.add_argument(
     "using Data Parallel or Distributed Data Parallel",
 )
 parser.add_argument(
-    "-eb",
-    "--effective-batch-size",
-    default=None,
+    "-na",
+    "--n_accumulation_steps",
+    default=1,
     type=int,
     metavar="N",
-    help="effective batch size used for SGD. Defaults to the "
-    "size of a mini-batch, but can be set to any integer "
-    "multiple of the mini-batch size",
+    help="number of mini-batches to accumulate into an effective batch for SGD",
 )
 parser.add_argument(
     "--lr",
@@ -224,15 +222,6 @@ def main():
         args.world_size = int(os.environ["WORLD_SIZE"])
 
     args.distributed = args.world_size > 1 or args.multiprocessing_distributed
-
-    if args.effective_batch_size is None:
-        args.effective_batch_size = args.batch_size
-
-    if args.effective_batch_size % args.batch_size != 0:
-        raise ValueError(
-            f"'effective_batch_size' ({args.effective_batch_size}) "
-            f"must be a multiple of 'batch_size' ({args.batch_size})"
-        )
 
     ngpus_per_node = torch.cuda.device_count()
     if args.multiprocessing_distributed:
@@ -389,7 +378,7 @@ def main_worker(gpu, ngpus_per_node, args):
         print("PRIVACY ENGINE ON")
         privacy_engine = PrivacyEngine(
             model,
-            batch_size=args.effective_batch_size,
+            batch_size=args.batch_size * args.n_accumulation_steps,
             sample_size=len(train_dataset),
             alphas=[1 + x / 10.0 for x in range(1, 100)] + list(range(12, 64)),
             noise_multiplier=args.sigma,
@@ -450,9 +439,6 @@ def train(train_loader, model, criterion, optimizer, epoch, args):
     # switch to train mode
     model.train()
 
-    # number of mini-batches to compute gradients on, before an actual update step
-    n_virtual_steps = args.effective_batch_size / args.batch_size
-
     end = time.time()
     for i, (images, target) in enumerate(train_loader):
         # measure data loading time
@@ -477,12 +463,12 @@ def train(train_loader, model, criterion, optimizer, epoch, args):
         optimizer.zero_grad()
         loss.backward()
 
-        if n_virtual_steps > 1:
+        if args.n_accumulation_steps > 1:
             optimizer.virtual_step()
 
         # make sure we take a step after processing the last mini-batch in the
         # epoch to ensure we start the next epoch with a clean state
-        if ((i + 1) % n_virtual_steps == 0) or ((i + 1) == len(train_loader)):
+        if ((i + 1) % args.n_accumulation_steps == 0) or ((i + 1) == len(train_loader)):
             optimizer.step()
 
         # measure elapsed time
--- a/examples/imagenet.py
+++ b/examples/imagenet.py
@@ -3,7 +3,6 @@
 
 """
 Runs ImageNet training with differential privacy.
-
 """
 
 import argparse
@@ -451,7 +450,7 @@ def train(train_loader, model, criterion, optimizer, epoch, args):
     optimizer.zero_grad()
 
     # number of mini-batches to compute gradients on, before an actual update step
-    n_accumulation_steps = args.effective_batch_size / args.batch_size
+    n_virtual_steps = args.effective_batch_size / args.batch_size
 
     end = time.time()
     for i, (images, target) in enumerate(train_loader):
@@ -476,10 +475,10 @@ def train(train_loader, model, criterion, optimizer, epoch, args):
         # compute gradient and do SGD step
         loss.backward()
 
-        if n_accumulation_steps > 1:
-            optimizer.accumulate_grads()
+        if n_virtual_steps > 1:
+            optimizer.virtual_step()
 
-        if (i + 1) % n_accumulation_steps == 0:
+        if (i + 1) % n_virtual_steps == 0:
             optimizer.step()
             optimizer.zero_grad()
 
--- a/examples/imagenet.py
+++ b/examples/imagenet.py
@@ -453,10 +453,13 @@ def train(train_loader, model, criterion, optimizer, epoch, args):
         top5.update(acc5[0], images.size(0))
 
         # compute gradient and do SGD step
+        optimizer.zero_grad()
         loss.backward()
-        if i % 4 == 3:
-            optimizer.step()
-            optimizer.zero_grad()
+        optimizer.step()
+
+        #if (i+1) % 2 == 0:
+        #    optimizer.step()
+        #    optimizer.zero_grad()
 
         # measure elapsed time
         batch_time.update(time.time() - end)
--- a/examples/imagenet.py
+++ b/examples/imagenet.py
@@ -447,7 +447,6 @@ def train(train_loader, model, criterion, optimizer, epoch, args):
 
     # switch to train mode
     model.train()
-    optimizer.zero_grad()
 
     # number of mini-batches to compute gradients on, before an actual update step
     n_virtual_steps = args.effective_batch_size / args.batch_size
@@ -473,14 +472,16 @@ def train(train_loader, model, criterion, optimizer, epoch, args):
         top5.update(acc5[0], images.size(0))
 
         # compute gradient and do SGD step
+        optimizer.zero_grad()
         loss.backward()
 
         if n_virtual_steps > 1:
             optimizer.virtual_step()
 
-        if (i + 1) % n_virtual_steps == 0:
+        # make sure we take a step after processing the last mini-batch in the
+        # epoch to ensure we start the next epoch with a clean state
+        if ((i + 1) % n_virtual_steps == 0) or ((i + 1) == len(train_loader)):
             optimizer.step()
-            optimizer.zero_grad()
 
         # measure elapsed time
         batch_time.update(time.time() - end)
--- a/examples/imagenet.py
+++ b/examples/imagenet.py
@@ -21,35 +21,34 @@ import torch.nn as nn
 import torch.nn.parallel
 import torch.optim
 import torch.utils.data
-import torch.utils.tensorboard as tensorboard
 import torch.utils.data.distributed
+import torch.utils.tensorboard as tensorboard
 import torchvision.datasets as datasets
 import torchvision.models as models
 import torchvision.transforms as transforms
-from torchdp import PrivacyEngine, utils
-from torchdp import stats
+from torchdp import PrivacyEngine, stats, utils
 
 
 # The following few lines, enable stats gathering about the run
 # 1. where the stats should be logged
-stats.set_global_summary_writer(
-    tensorboard.SummaryWriter('/tmp/stats'))
+stats.set_global_summary_writer(tensorboard.SummaryWriter("/tmp/stat"))
 # 2. enable stats
 stats.add(
     # stats about gradient norms aggregated for all layers
-    stats.Stat(
-        stats.StatType.CLIPPING, 'AllLayers', frequency=0.1),
+    stats.Stat(stats.StatType.CLIPPING, "AllLayers", frequency=0.1),
     # stats about gradient norms per layer
-    stats.Stat(
-        stats.StatType.CLIPPING, 'IndividualLayers', frequency=0.1),
+    stats.Stat(stats.StatType.CLIPPING, "PerLayer", frequency=0.1),
+    # stats about clipping
+    stats.Stat(stats.StatType.CLIPPING, "ClippingStats", frequency=0.1),
     # stats on training accuracy
-    stats.Stat(
-        stats.StatType.TRAIN, 'accuracy', frequency=0.01),
+    stats.Stat(stats.StatType.TRAIN, "accuracy", frequency=0.01),
     # stats on validation accuracy
-    stats.Stat(
-        stats.StatType.TEST, 'accuracy'),
+    stats.Stat(stats.StatType.TEST, "accuracy"),
 )
 
+# The following lines enable stat gathering for the clipping process
+# and set a default of per layer clipping for the Privacy Engine
+clipping = {"clip_per_layer": True, "enable_stat": True}
 
 parser = argparse.ArgumentParser(description="PyTorch ImageNet DP Training")
 parser.add_argument("data", metavar="DIR", help="path to dataset")
@@ -377,6 +376,7 @@ def main_worker(gpu, ngpus_per_node, args):
             alphas=[1 + x / 10.0 for x in range(1, 100)] + list(range(12, 64)),
             noise_multiplier=args.sigma,
             max_grad_norm=args.max_per_sample_grad_norm,
+            **clipping,
         )
         privacy_engine.attach(optimizer)
     else:
--- a/examples/imagenet.py
+++ b/examples/imagenet.py
@@ -47,7 +47,7 @@ stats.add(
 
 # The following lines enable stat gathering for the clipping process
 # and set a default of per layer clipping for the Privacy Engine
-clipping = {"clip_per_layer": True, "enable_stat": True}
+clipping = {"clip_per_layer": False, "enable_stat": True}
 
 parser = argparse.ArgumentParser(description="PyTorch ImageNet DP Training")
 parser.add_argument("data", metavar="DIR", help="path to dataset")
--- a/examples/imagenet.py
+++ b/examples/imagenet.py
@@ -229,8 +229,10 @@ def main():
         args.effective_batch_size = args.batch_size
 
     if args.effective_batch_size % args.batch_size != 0:
-        raise ValueError(f"'effective_batch_size' ({args.effective_batch_size}) "
-                         f"must be a multiple of 'batch_size' ({args.batch_size})")
+        raise ValueError(
+            f"'effective_batch_size' ({args.effective_batch_size}) "
+            f"must be a multiple of 'batch_size' ({args.batch_size})"
+        )
 
     ngpus_per_node = torch.cuda.device_count()
     if args.multiprocessing_distributed:
--- a/examples/imagenet.py
+++ b/examples/imagenet.py
@@ -230,7 +230,8 @@ def main():
         args.effective_batch_size = args.batch_size
 
     if args.effective_batch_size % args.batch_size != 0:
-        raise ValueError("'effective_batch_size' must be a multiple of 'batch_size'")
+        raise ValueError(f"'effective_batch_size' ({args.effective_batch_size}) "
+                         f"must be a multiple of 'batch_size' ({args.batch_size})")
 
     ngpus_per_node = torch.cuda.device_count()
     if args.multiprocessing_distributed:
--- a/examples/imagenet.py
+++ b/examples/imagenet.py
@@ -372,7 +372,7 @@ def main_worker(gpu, ngpus_per_node, args):
         print("PRIVACY ENGINE ON")
         privacy_engine = PrivacyEngine(
             model,
-            batch_size=args.batch_size,
+            batch_size=args.batch_size * 4,
             sample_size=len(train_dataset),
             alphas=[1 + x / 10.0 for x in range(1, 100)] + list(range(12, 64)),
             noise_multiplier=args.sigma,
@@ -453,9 +453,10 @@ def train(train_loader, model, criterion, optimizer, epoch, args):
         top5.update(acc5[0], images.size(0))
 
         # compute gradient and do SGD step
-        optimizer.zero_grad()
         loss.backward()
-        optimizer.step()
+        if i % 4 == 3:
+            optimizer.step()
+            optimizer.zero_grad()
 
         # measure elapsed time
         batch_time.update(time.time() - end)
--- a/examples/imagenet.py
+++ b/examples/imagenet.py
@@ -81,6 +81,16 @@ parser.add_argument(
     "using Data Parallel or Distributed Data Parallel",
 )
 parser.add_argument(
+    "-eb",
+    "--effective-batch-size",
+    default=None,
+    type=int,
+    metavar="N",
+    help="effective batch size used for SGD. Defaults to the "
+         "size of a mini-batch, but can be set to any integer "
+         "multiple of the mini-batch size",
+)
+parser.add_argument(
     "--lr",
     "--learning-rate",
     default=0.1,
@@ -215,6 +225,12 @@ def main():
         args.world_size = int(os.environ["WORLD_SIZE"])
 
     args.distributed = args.world_size > 1 or args.multiprocessing_distributed
+
+    if args.effective_batch_size is None:
+        args.effective_batch_size = args.batch_size
+
+    if args.effective_batch_size % args.batch_size != 0:
+        raise ValueError("'effective_batch_size' must be a multiple of 'batch_size'")
 
     ngpus_per_node = torch.cuda.device_count()
     if args.multiprocessing_distributed:
@@ -371,7 +387,7 @@ def main_worker(gpu, ngpus_per_node, args):
         print("PRIVACY ENGINE ON")
         privacy_engine = PrivacyEngine(
             model,
-            batch_size=args.batch_size * 4,
+            batch_size=args.effective_batch_size,
             sample_size=len(train_dataset),
             alphas=[1 + x / 10.0 for x in range(1, 100)] + list(range(12, 64)),
             noise_multiplier=args.sigma,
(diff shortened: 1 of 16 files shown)


torchdp/autograd_grad_sample.py
```diff
@@ -31,29 +31,39 @@
 # work-around for https://github.com/pytorch/pytorch/issues/25723
 _hooks_disabled: bool = False
 
-# global switch to catch double backprop errors on Hessian computation
-_enforce_fresh_backprop: bool = False
 
-
-def add_hooks(model: nn.Module) -> None:
+def add_hooks(model: nn.Module, loss_type: str = "mean", batch_dim: int = 0) -> None:
```
Anonymani, line 35: nit: we could pass an IntEnum instead.

torchdp/autograd_grad_sample.py
```diff
@@ -31,29 +31,39 @@
...
     handles = []
     for layer in model.modules():
         if get_layer_type(layer) in _supported_layers:
             handles.append(layer.register_forward_hook(_capture_activations))
-            handles.append(layer.register_backward_hook(_capture_backprops))
+
+            def backward_hook_closure(layer, _input, output):
+                _capture_backprops(layer, _input, output, loss_type, batch_dim)
+
+            handles.append(layer.register_backward_hook(backward_hook_closure))
```
Anonymani, line 66: nit:
handles.append(layer.register_backward_hook(partial(backward_hook_closure, loss_type=loss_type, batch_dim=batch_dim))

torchdp/autograd_grad_sample.py
ftramer: Some parts of the code (e.g., parts of the clipper) silently assume that loss_type=mean, and batch_dim=0. 
Is this generality really warranted here? E.g., I don't see a compelling reason why the batch dimension wouldn't be zero.

torchdp/autograd_grad_sample.py
ftramer: done, thanks!

torchdp/autograd_grad_sample.py
```diff
@@ -105,136 +115,160 @@ def _capture_activations(
...
-            )
-        _enforce_fresh_backprop = False
-
-    if not hasattr(layer, "backprops_list"):
-        layer.backprops_list = []
-    layer.backprops_list.append(output[0].detach())
+    backprops = output[0].detach()
+    _compute_grad_sample(layer, backprops, loss_type, batch_dim)
+    if hasattr(layer, "activations"):
+        del layer.activations
```
Anonymani, line 137: nit: IMO, at one point in time we are going to have all activations, back-props, and grad_samples from one micro batch in memory, so really there is no point in deleting these unless we want to make space for other programs which we are not optimizing here for. let's just leave them be. deleting memory and recreating it only harms the perf.

ftramer: agreed

torchdp/per_sample_gradient_clip.py
```diff
@@ -171,7 +188,45 @@ def clip(self) -> List[float]:
...
-        return threshs
+
+        if virtual:
+            # check if we're adding to an existing accumulator
+            if "clip_threshs" in self.accumulation_state:
+                # retain the largest clipping thresholds accross the entire batch
+                curr_thresh = self.accumulation_state["clip_threshs"]
+                self.accumulation_state["clip_threshs"] = [
+                    max(n1, n2) for (n1, n2) in zip(threshs, curr_thresh)
+                ]
```
Anonymani, line 199: very neat! May be, I'd also leave the option of not changing the thresh and using the first batch thresh for the rest.  The reason being if someone uses a very small mini-batch norm, this could result in the a very large thresh to be used, for this diff may be you could simply add a TODO or warning here and leave it as is.

torchdp/per_sample_gradient_clip.py
```diff
@@ -195,22 +250,35 @@ def __del__(self):
...
+        """
+        Clips and sums up per-sample gradients into an accumulator.
+        After calling self.virtual_step() N times on mini-batches of
+        B per-sample gradients, a call to self.step() will 
+        populate the .grad field with the average gradient over
+        the entire batch of size N*B.
+        """
+        self.gradient_clipper.clip(virtual=True)
+        # reset the accumulation of per-sample gradients
+        autograd_grad_sample.clear_grad_sample(self.module)
```
Anonymani, line 276: this kind of goes against having a create_or_extend function unless we allow extension. may be adding a mode like accumulating : bool or sth similar could help. I am not suggesting that we do extension in this diff but knowing it exists help. This way we do not have to delete the grad_sample when simply only accumulating.

torchdp/autograd_grad_sample.py
```diff
@@ -105,136 +115,160 @@ def _capture_activations(
...
+def _create_or_extend_grad_sample(
+    param: torch.Tensor, grad_sample: torch.Tensor, batch_dim: int
+) -> None:
+    """
+    Create a 'grad_sample' attribute in the given parameter, or append to it
+    if it already exists.
+    """
+
+    if hasattr(param, "grad_sample"):
+        param.grad_sample = torch.cat((param.grad_sample, grad_sample), batch_dim)
```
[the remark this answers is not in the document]
ftramer: This gets a little annoying because now param.grad_sample can be either a tensor or a list. Then, when we call .clip(), before we use .grad_sample we'd have to first check whether it's a list and if so call torch.cat(). Definitely doable though.

torchdp/per_sample_gradient_clip.py
```diff
@@ -125,39 +126,55 @@ def calc_thresh_value(self, layer_norms):
...
+            virtual (bool): if set, the clipped gradients in this mini-batch
+            are summed up into a accummulator for a larger batch
 
         Returns:
-            A dictionary of parameter names and their clipped norm
+            A list of clipping thresholds
         """
+
+        # check if we've already accumulated all the clipped gradients for this batch
+        if self.accumulation_state and not virtual:
```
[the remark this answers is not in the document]
ftramer: done

torchdp/autograd_grad_sample.py
```diff
@@ -105,136 +115,160 @@ def _capture_activations(
...
 
 
-def clear_backprops(model: nn.Module) -> None:
-    """Delete layer.backprops_list in every layer."""
+def clear_grad_sample(model: nn.Module) -> None:
+    """Delete 'param.grad_sample' in every parameter of the model."""
     for layer in model.modules():
-        if hasattr(layer, "backprops_list"):
-            del layer.backprops_list
+        for param in layer.parameters():
```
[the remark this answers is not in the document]
ftramer: done

torchdp/per_sample_gradient_clip.py
```diff
@@ -171,7 +188,45 @@ def clip(self) -> List[float]:
...
+        stores them in the .grad field
+        """
+
+        batch_size = self.accumulation_state["batch_size"]
+        # now that we know the full batch size, we can average the gradients
+        for _, p in self.named_params:
+            acc_grad = self.accumulation_state[p]
+            p.grad = acc_grad / batch_size
+
+        threshs = self.accumulation_state["clip_threshs"].copy()
```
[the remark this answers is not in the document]
ftramer: Right, I'm resetting the dictionary in erase_virtual_batch but this won't affect the list.
Fixed

torchdp/per_sample_gradient_clip.py
ftramer: agreed. I added a TODO.

torchdp/per_sample_gradient_clip.py
```diff
@@ -125,39 +126,55 @@ def calc_thresh_value(self, layer_norms):
...
+            summed_grad = torch.einsum("i,i...", per_sample_clip_factor, p.grad_sample)
+
+            if virtual:
+                # accumulate the summed gradient for this mini-batch
+                if p in self.accumulation_state:
+                    self.accumulation_state[p] += summed_grad
+                else:
+                    self.accumulation_state[p] = summed_grad
+            else:
+                p.grad = summed_grad / batch_size
```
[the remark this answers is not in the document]
ftramer: done

torchdp/per_sample_gradient_clip.py
ftramer: Note that accumulating gradients in .grad_sample via create_or_extend doesn't require any calls to .virtual_step(). This happens automatically via the backprop hooks.
So my thoughts here were that you can either call .backward() multiple times to have the per-sample gradients accumulate in .grad_sample, or clip and sum up gradients into .summed_grad by calling .virtual_step().
You can also combine these two (there's an example in virtual_step_test.py that does this).

But after you've called .virtual_step(), it is important to clear .grad_sample since those gradients have all been clipped and summed up already, and we wouldn't want to process them again.

torchdp/autograd_grad_sample.py
Anonymani: you are right, it is probably unnecessary  but let's do that in another diff.

ftramer: Ok! then I'll leave this as is for now

torchdp/autograd_grad_sample.py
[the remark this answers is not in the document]
Anonymani: we could have grad_sample_list for that purpose. Anyhow, Since this is not happening too often (we might extend at most 4 times and the do accumulation) I am not insisting on it as the perf gain is probably negligible.

torchdp/per_sample_gradient_clip.py
```diff
@@ -125,39 +126,55 @@ def calc_thresh_value(self, layer_norms):
...
             # We are *clipping* the gradient, so if the factor is ever >1 we set it to 1
             per_sample_clip_factor = per_sample_clip_factor.clamp(max=1.0)
-            b_sz = len(per_sample_clip_factor)  # all batch sizes are the same
             # step 2: Do the clipping
             name, p = named_param
             pre_clip_pos = p.grad_sample.mean(0) > 0
-            p.grad = (
-                torch.einsum("i,i...", per_sample_clip_factor, p.grad_sample) / b_sz
-            )
+            summed_grad = torch.einsum("i,i...", per_sample_clip_factor, p.grad_sample)
```
Anonymani, line 167: Do we need p.grad_sample after this line anymore?
Could we remove the clear_grad_sample function and do this here?

ftramer: many tests in layers_grad_test.py assume that p.grad_sample will still exist after we've called clipper.step(). So at the very least this would require changing all those tests.

Also, we'd still need clear_grad_sample() because it has to be called in optimizer.zero_grad(): if the effective batch size doesn't evenly divide the training set size, the last effective batch in a training epoch is incomplete: we accumulate a bunch of minibatches, but never call .step()
By calling .zero_grad() at the start of an epoch, we ensure that all the accumulated gradients are deleted.

torchdp/per_sample_gradient_clip.py
```diff
@@ -171,7 +188,54 @@ def clip(self) -> List[float]:
...
+        """
+        Averages the clipped gradients aggregated over multiple mini-batches and
+        stores them in the .grad field
+        """
+
+        batch_size = self.accumulation_state["batch_size"]
+        # now that we know the full batch size, we can average the gradients
+        for _, p in self.named_params:
+            acc_grad = p.summed_grad
+            p.grad = acc_grad / batch_size
```
Anonymani, line 225: same here,  from the looks of it we can just discard p.summed-grad here

torchdp/per_sample_gradient_clip.py
```diff
@@ -171,7 +188,54 @@ def clip(self) -> List[float]:
...
+
+        threshs = self.accumulation_state["clip_threshs"]
+        self.erase_virtual_batch()
```
Anonymani, line 228: following comments above:
we can simply have self.accumulation_state = {}

torchdp/per_sample_gradient_clip.py
Anonymani: also some comments below on some other cleanings.

torchdp/per_sample_gradient_clip.py
ftramer: This happens in the call to erase_virtual_batch 2 lines below.
I extracted this into its own methos so that it can also be called in .zero_grad(), to address the same concern as above, namely that we could finish a training epoch with an incomplete effective batch.

torchdp/per_sample_gradient_clip.py
Anonymani: and then remove erase_virtual_batch function

ftramer: we do need that function in .zero_grad() to be able to erase incomplete effective batches.
Let me know if you're thinking of a better way to handle this.

examples/imagenet.py
```diff
@@ -453,9 +473,14 @@ def train(train_loader, model, criterion, optimizer, epoch, args):
...
         # compute gradient and do SGD step
-        optimizer.zero_grad()
         loss.backward()
-        optimizer.step()
+
+        if n_virtual_steps > 1:
+            optimizer.virtual_step()
+
+        if (i + 1) % n_virtual_steps == 0:
+            optimizer.step()
```
Anonymani, line 482: are we not taking a step if len_data/batch is not dividable by effective_batch

ftramer: Here's an example of what I had in mind:
- len(train_loader) = 1250
- batch_size = 100
- effective_batch_size = 500

so n_virtual_steps = 5, and i ranges from 0 to 12

In a training epoch, we'll call .step() when (i+1)=5, and when (i+1)=10, to process batches of size 500.
Then, the last steps will process mini-batches of size 100, 100, 50, and exit the training loop.
So we're left with 3 mini-batches accumulated into param.grad_sample or param.summed_grad

Anonymani: exactly, so what's happening to those in the next epoch? am I missing something?
we either have to miss them by calling zero grad or handle them at the end of the loop, correct?

ftramer: right, so in the next epoch, `i` gets reset to 0 so we'll accumulate 5 more mini-batches before we call `optimizer.step()`.
So now we have an effective batch consisting of 8 mini-batches (3 from the previous epoch, and 5 from the current epoch) and with 750 samples total. 

This is problematic for two reasons: 

1. It could cause memory issues if we're planning on holding the entire batch in memory, and we were expecting at most 500 gradients.

2. The privacy accounting for that batch is incorrect because we're assuming a sample probability of `effective_batch_size / len(train_loader) = 500/1250` whereas this batch has a sample probability of `750/1250`.

Anonymani: So my thought is we should err on the conservative side (as epsilon is a an upper-bound). 
To do that I'd suggest:
either:
calling step() once at the end of the loop and checking inside step if there is un-finished business, we could check the accumulation_state or yet cleaner have a flag that tracks state. if a virtual step is called the flag is set and on virtual=false flag is cleared.
or:
do the same, but call the method finalize() [just sounds better and less confusing IMO]. For this you could simply just promote finalize_batch.
or:
handle everything internally in the engine counting number of calls to step and finalizing whenever needed. For this there would be no need for a virtual_step method at all. [I kind of like this one, but also I know this completely contradicts the talks we had offline, so may be we can do this in another diff later on [let me know what you think about this :) ]

torchdp/privacy_engine.py
```diff
@@ -95,16 +116,33 @@ def get_privacy_spent(self, target_delta: float = None):
...
     def step(self):
         self.steps += 1
-        clip_values = self.clipper.step()
+        clip_values, batch_size = self.clipper.step()
+
+        # ensure the clipper consumed the right amount of gradients.
+        # In the last batch of a training epoch, we might get a batch that is
+        # smaller than others but we should never get a batch that is too large
+        if batch_size > self.batch_size:
+            raise ValueError(
```
Anonymani, line 125: I like this ValueError, but this contradicts what you mentioned earlier in the comments on imagenet.py. I left some comments there. I would may be also add a Warning for batch_size < self.batch_size which just prints out about the potential non-optimal accumulation in the last rounds of the loop.

ftramer: This ValueError is not raised in the imagenet.py example because we're calling `optimizer.zero_grad()` at the beginning of each epoch. So any accumulated, but not yet finalized gradients get erased.
I'll add the warning.

examples/imagenet.py
ftramer: I guess the easiest way would be to do:
```python
if ((i + 1) % n_virtual_steps == 0) or ((i+1) == len(train_loader)):
    optimizer.step()
```

Then we'll always take a step after the last mini-batch, regardless of whether we've completed an effective batch or not.

Anonymani: I like that!

torchdp/privacy_engine.py
Anonymani: Cool

examples/imagenet.py
Anonymani: could you make this change as well please (I'd assume it would need some minimal changes in the step function as well) and we should be more than ready to land this amazing diff.

ftramer: Ok. There's still the question of when/where to delete the `param.grad_sample` fields so that they stop accumulating.
If we want to do that in `clipper.step()`, we'd have to rewrite a bunch of tests.
I think the most natural thing is still to monkeypatch `optimizer.zero_grad()` and do this there, as it aligns nicely with the semantics of `zero_grad()` for accumulation of regular gradients.

ftramer: With the latest commits, the deletions are now handled directly in `clipper.step()`, so that gradient aggregators are reset as soon as they are consumed.
And we just have to make sure to always take a step on the last mini-batch of a training epoch.

torchdp/autograd_grad_sample.py
```diff
@@ -105,136 +119,150 @@ def _capture_activations(
...
-            del layer.backprops_list
+    if hasattr(param, "grad_sample"):
+        param.grad_sample = torch.cat((param.grad_sample, grad_sample), batch_dim)
+    else:
+        param.grad_sample = grad_sample
 
 
-def compute_grad_sample(
-    model: nn.Module, loss_type: str = "mean", batch_dim: int = 0
+def _compute_grad_sample(
```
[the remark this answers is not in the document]
ftramer: I messed up something in my fork so ended up pulling and merging rather than rebasing...
Hope this works!

examples/imagenet.py
```diff
@@ -80,6 +79,16 @@
     "batch size of all GPUs on the current node when "
     "using Data Parallel or Distributed Data Parallel",
 )
+parser.add_argument(
+    "-eb",
+    "--effective-batch-size",
```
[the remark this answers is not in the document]
ftramer: Done

torchdp/privacy_engine.py
```diff
@@ -95,16 +111,36 @@ def get_privacy_spent(self, target_delta: float = None):
...
+                f"PrivacyEngine expected a batch of size {self.batch_size} "
+                f"but received a batch of size {batch_size}"
+            )
+
+        if batch_size < self.batch_size:
+            warnings.warn(
+                f"PrivacyEngine expected a batch of size {self.batch_size} "
+                f"but received a batch of size {batch_size}. The privacy "
```
Darktex, line 128: ... But the last step had batch_size = {batch_size}. This means we will be a bit more pessimistic in our analysis. Consider setting `drop_last = True` in your PyTorch dataloader to avoid this problem completely

Anonymani:
@ftramer a bunch of automatic errors that is blocking push:

torchdp/test/privacy_engine_test.py
Line 242
Loop control variable 'layer_name' not used within the loop body. If this is intended, start the name with an underscore.

torchdp/test/virtual_step_test.py
Line 125
Loop control variable 'idx' not used within the loop body. If this is intended, start the name with an underscore.
Line 164
Loop control variable 'idx' not used within the loop body. If this is intended, start the name with an underscore.
Line 302
Loop control variable 'i' not used within the loop body. If this is intended, start the name with an underscore.

torchdp/per_sample_gradient_clip.py
Line 276
trailing whitespace

torchdp/test/virtual_step_test.py
Line 286
trailing whitespace

torchdp/test/privacy_engine_test.py
```diff
@@ -163,8 +161,8 @@ def test_throws_double_attach(self):
         model, optimizer = self.setUp_init_model(private=True)
         self.setUp_model_step(model, optimizer)
         with self.assertRaises(ValueError):
-            model, optimizer = self.setUp_init_model(private=True, model=model)
-            self.setUp_model_step(model, optimizer)
+            model, optimizer2 = self.setUp_init_model(private=True, model=model)
```
karthikprasad, line 164: was there a bug in the test before this change?

torchdp/per_sample_gradient_clip.py
```diff
@@ -125,39 +126,58 @@ def calc_thresh_value(self, layer_norms):
             thresh_norm if self.clip_per_layer else thresh_norm * len(self.named_params)
         )
 
-    def clip(self) -> List[float]:
+    def clip(self, virtual: bool = False) -> (List[float], int):
```
karthikprasad, line 129: nit: change bool arg name to `is_virtual`?

torchdp/per_sample_gradient_clip.py
```diff
@@ -125,39 +126,58 @@ def calc_thresh_value(self, layer_norms):
...
         We assume the batch size is the first dimension.
 
         Arguments:
-            tensor (Tensor): a single Tensor whose norm will be normalized
+            virtual (bool): if set, the clipped gradients in this mini-batch
+            are summed up into a accummulator for a larger batch
 
         Returns:
-            A dictionary of parameter names and their clipped norm
+            A list of clipping thresholds
```
karthikprasad, line 146: this function reports a tuple now.

ftramer: Fixed

torchdp/per_sample_gradient_clip.py
ftramer: Fixed

torchdp/test/privacy_engine_test.py
ftramer: Good catch! This test had to be changed in a previous design of virtual batches. With the current design, the original test passes.

torchdp/privacy_engine.py
ftramer: Done. Thanks

ftramer:
@Anonymani the "trailing whitespace" errors seem to be in comments so I'm not quite sure what the error is there. Are they fixed now?

Anonymani:
> @Anonymani the "trailing whitespace" errors seem to be in comments so I'm not quite sure what the error is there. Are they fixed now?

let me re-import and see report errors [if any] again.

torchdp/test/virtual_step_test.py
```diff
@@ -0,0 +1,311 @@
...
+            loss.backward()
+
+        self.effective_batch_grad = torch.cat(
+            [p.grad.reshape(-1) for p in self.model.parameters() if p.requires_grad]
+        ) * (self.BATCH_SIZE / self.DATA_SIZE)
+
+        self.optimizer.zero_grad()
+
+        # accumulate grads for each pair of batches
+        self.double_grads = []
```
ashkan-software, line 84: Could you please add a comment in places that you use `double_grads`, in that what is the "pair of batches" and why we have a pair?

torchdp/test/virtual_step_test.py
```diff
@@ -0,0 +1,311 @@
...
+        self.assertTrue(torch.allclose(grad, orig_grad, atol=10e-5, rtol=10e-3))
+
+    def test_zero_grad(self):
+        """
+        Calling optimizer.zero_grad() should erase any accumulated per-sample gradients.
+        """
+        self.setUp_privacy_engine(2 * self.BATCH_SIZE)
+
+        double_grads_per_sample = []
+        double_grads = []
```
ashkan-software, line 235: The same comment I put for commenting `double_grads` :)

torchdp/test/virtual_step_test.py
```diff
@@ -0,0 +1,311 @@
...
+        double_grads = torch.cat(double_grads)
+        double_grads_per_sample = torch.cat(double_grads_per_sample)
+
+        self.assertTrue(
+            torch.allclose(
+                self.double_grads, double_grads_per_sample, atol=10e-5, rtol=10e-3
+            )
+        )
+        self.assertTrue(
+            torch.allclose(self.double_grads, double_grads, atol=10e-5, rtol=10e-3)
```
ashkan-software, line 281: Hmm I am now a bit confused by this. :) We are checking if `self.double_grads` is close to `double_grads`. Should we use a better naming for these variables? (comparing `x` and `self.x` may not be very clear)

torchdp/test/virtual_step_test.py
ftramer: Sorry, these were a relic for a test of the behavior of zero_grad(), which we ultimately changed. I've refactored the test to check that accumulated gradients are correctly erased upon a call to `.step()`. These "double_grads" are no longer needed.

torchdp/test/virtual_step_test.py
ftramer: as above

torchdp/test/virtual_step_test.py
ftramer: as above

PR #30, opened 2020-08-02
Add implementation of Gaussian DP

woodyx218:
## Types of changes

- [X] New feature (non-breaking change which adds functionality)

Add privacy accountants for Gaussian Differential Privacy (GDP) from my paper https://arxiv.org/abs/1911.11607 as "compute_GDP_privacy.py". The accountants calculate GDP of many optimizers with uniform or Poisson subsampling.

Add one tutorial for Adult Income Dataset, which is a classic tabular binary-classification task. It corresponds to the experiment from the same paper. This will be a good complement to the image classification tasks pytorch-dp have as examples.

The code has been tested on GPU locally and passed the formatting (flake8) test 

## Checklist

- [X] The documentation is up-to-date with the changes I made.
- [X] I have read the **CONTRIBUTING** document and completed the CLA (see **CONTRIBUTING**).
- [X] All tests passed, and additional code has been covered with new tests.

### DIFF
--- a/examples/adult_income.py
+++ b/examples/adult_income.py
@@ -193,6 +193,7 @@ def main():
 
     """Loads ADULT a2a as in LIBSVM and preprocesses to combine training and validation data."""
     # https://www.csie.ntu.edu.tw/~cjlin/libsvmtools/datasets/binary.html
+    # can also find in github.com/woodyx218/Deep-Learning-with-GDP-Pytorch/blob/master/adult.csv
     x = pd.read_csv("adult.csv")
     trainData, testData = train_test_split(x, test_size=0.1, random_state=218)
     # have to reset index, see https://discuss.pytorch.org/t/keyerror-when-enumerating-over-dataloader/54210/13
--- a/examples/adult_income.py
+++ b/examples/adult_income.py
@@ -1,0 +1,253 @@
+"""
+Runs Adult training with differential privacy.
+"""
+
+import argparse
+
+import numpy as np
+import pandas as pd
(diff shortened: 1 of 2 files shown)


woodyx218:
can anyone take a look?

Darktex:
Hi @woodyx218 ! 

First of all, sorry for the late reply! We are hard at work with some refactoring and documenting so this slipped for a bit :(

We are truly thankful for this PR, but we would rather this stayed outside for the time being - let me share why. Our aim is to provide components that people can use to either train models with DP, or to do research in the field - kinda like `torch.nn`. A core responsibility of ours is to maintain every line of code in the repo from today to infinity, so in order to balance our time we chose to commit to supporting only features that have already very high demand. This comes with its own set of problems of course - for example, Transformers were published in 2017 but `nn.Transformer` only showed up in 2019. Luckily, community contributions such as HuggingFace's package bridged the gap so most PyTorch users were able to train transformers in PyTorch long before `nn.Transformer` was done. We aim to have a similar model here.

That being said, we are looking for ways to give researchers using PyTorch-DP more visibility to the community, so that we can all share ideas better. We are still in the process of defining this (and we are writing a proper contributor manual to state this more clearly), but what we have in mind is to list papers&code somewhere, kinda like what PyTorch does with its [ecosystem](https://pytorch.org/ecosystem/).

We would be glad to link your paper & repo https://github.com/woodyx218/Deep-Learning-with-GDP-Pytorch in there and when this becomes a new accountant that many projects depend on, you can count on us adding first-class support for it :)

PR #56, opened 2020-09-12
Add GradSampleModule

Darktex:
Summary: Having grad samples is a core abstraction in DP-SGD. The way we currently handle grad samples is a bit magical with hooks, global variables and all. This introduces a more logical reorganization of our code by promoting grad sample hooks to be not just that the privacy engine attaches at some point, but a core abstraction and concept, in line with the theory.

Differential Revision: D22991710


Darktex:
PR abandoned in favor of #149

PR #82, opened 2020-11-04
run Transformer example test on torch nightly (#81)

Darktex:
Summary:
Pull Request resolved: https://github.com/pytorch/opacus/pull/81

The imdb example using Transformer is using dataset from torchtext experimental, which is a prototype and not available in stable release builds (see https://github.com/pytorch/text/releases/tag/v0.8.0-rc2)

This is causing circle ci failures. This commit runs that specific test on torch nightly. This separate run can be removed once the "Prototype" gets promoted to "Beta"

Differential Revision: D24677803


Darktex:
Split into two PRs

Issue #103, opened 2020-12-11
Extend opacus.DPLSTM to support PackedSequences


Darktex:
## 🚀 Feature
Extend [opacus.DPLSTM](https://github.com/pytorch/opacus/blob/master/opacus/layers/dp_lstm.py) to work with [PackedSequences](https://pytorch.org/docs/stable/generated/torch.nn.utils.rnn.PackedSequence.html).

This is a good first issue to contribute, and we would very much welcome a PR!

## Motivation

The `PackedSequence` format allows us to minimize padding in a batch by "zipping" sequences together, and keeping track of the lengths. It is a very commonly-used format for torch.nn.LSTM and it is the only feature that our DPLSTM reimplementation does not yet support. 

## Pitch

A common problem in NLP (or in dealing with sequences of stuff, more in general) is uneven lengths of examples in a batch. For example, if you want to batch `short sentence` and `very long long long long sentence` together in a single tensor, you need to handle the fact that the first sequence has length of 2 and the second has length of 6. Matrices (and tensors) can't have uneven rows :)

A common approach to this is simply padding: add a special token, eg `<pad>` to all sequences as needed to reach the length of the longest sequence in the batch (eg `short sentence <pad> <pad> <pad> <pad>`). This works, but leads to wasted memory and computation (the LSTM will still run through the pads). A better approach is having a format such as `PackedSequence`, which instead stacks and zips the sequences, and keeps track of all the relative lengths (see ***Additional Context** for more info on how this is done). This way, we don't waste neither memory nor compute. In our example before, we would pack our two sequences together like this:

`[short, very, sentence, long, long, long, long, sentence]` with `lens=[2, 6]`.

So we "zip" them, and then we just keep going once the shorter sequence is finished. Why do we zip them? Think about how the LSTM runs: it goes step by step, but you can batch every step! This way, we start by running on both `short` and `very` (first items in each seq), then we continue to `sentence` and `long`, and then we can return the result for the first sequence but we keep going for the second sequence (we know how to do this because we have `lens` to tell us).

## Additional context

See more tutorials about PackedSequences:

1. https://gist.github.com/HarshTrivedi/f4e7293e941b17d19058f6fb90ab0fec
2. https://discuss.pytorch.org/t/understanding-pack-padded-sequence-and-pad-packed-sequence/4099/6
3. This StackOverflow [answer](https://stackoverflow.com/a/56211056/371613) has a good visual explanation.

adriansarstedt:
[...]

PR #116, opened 2021-01-17
Added PackedSequence support in DPLSTM

touqir14:
## Types of changes

This is a pull request in response to the [issue](https://github.com/pytorch/opacus/issues/103) for adding support for PackedSequence in DPLSTM. An example usage is shown below:
```python
from opacus.layers import DPLSTM
import torch 

seq_batch = [torch.tensor([[1, 1],
                           [2, 2],
                           [3, 3],
                           [4, 4],
                           [5, 5]]),
             torch.tensor([[10, 10],
                           [20, 20]])]

seq_lens = [5, 2]

padded_seq_batch = torch.nn.utils.rnn.pad_sequence(seq_batch, batch_first=True)
packed_seq_batch = torch.nn.utils.rnn.pack_padded_sequence(padded_seq_batch, lengths=seq_lens, batch_first=True)

dp_lstm = DPLSTM(input_size=2, hidden_size=3, batch_first=True)
output, (hn, cn) = dp_lstm(packed_seq_batch.float())

```

This has been tested analogously to the way that DPLSTM (without PackedSequence) has been tested. The outputs and parameters of DPLSTM (with PackedSequence) and LSTM (with PackedSequence) have been compared in the tests.
## Checklist

### DIFF
--- a/opacus/autograd_grad_sample.py
+++ b/opacus/autograd_grad_sample.py
@@ -18,7 +18,6 @@ from typing import Tuple
 
 import torch
 import torch.nn as nn
-
 from opacus.layers.dp_lstm import LSTMLinear
 
 from .supported_layers_grad_samplers import _supported_layers_grad_samplers
--- a/opacus/autograd_grad_sample.py
+++ b/opacus/autograd_grad_sample.py
@@ -218,7 +218,10 @@ def _compute_grad_sample(
     else:
         A = layer.activations
 
-    n = A.shape[batch_dim]
+    if not hasattr(layer, 'max_batch_len'):
+        layer.max_batch_len = _get_batch_size(layer, A, batch_dim)
+
+    n = layer.max_batch_len
     if loss_reduction == "mean":
         B = backprops * n
     elif loss_reduction == "sum":
@@ -238,4 +241,24 @@ def _compute_grad_sample(
     )
 
     compute_layer_grad_sample(layer, A, B)        
+
+    if (not isinstance(layer.activations, list) or len(layer.activations) == 0) and hasattr(layer, 'max_batch_len'):
+        del layer.max_batch_len
+
+
+def _get_batch_size(layer: nn.Module, grad_sample: torch.Tensor, batch_dim: int) -> int:
+    r"""
+    Computes and returns the maximum batch size which is the maximum of the dimension values
+    along 'batch_dim' axis over layer.activations + [grad_sample], where layer.activations is 
+    a list. If layer.activations is a not a list, then return grad_sample.shape[batch_dim]. 
+    """
+    
+    max_batch_len = 0
+    if isinstance(layer.activations, list):
+        for out in layer.activations:
+            if out.shape[batch_dim] > max_batch_len:
+                max_batch_len = out.shape[batch_dim]
+
+    max_batch_len = max(max_batch_len, grad_sample.shape[batch_dim])
+    return max_batch_len
 
--- a/opacus/autograd_grad_sample.py
+++ b/opacus/autograd_grad_sample.py
@@ -218,7 +218,7 @@ def _compute_grad_sample(
     else:
         A = layer.activations
 
-    if not hasattr(layer, 'max_batch_len'):
+    if not hasattr(layer, "max_batch_len"):
         layer.max_batch_len = _get_batch_size(layer, A, batch_dim)
 
     n = layer.max_batch_len
@@ -240,19 +240,21 @@ def _compute_grad_sample(
         get_layer_type(layer)
     )
 
-    compute_layer_grad_sample(layer, A, B)        
+    compute_layer_grad_sample(layer, A, B)
 
-    if (not isinstance(layer.activations, list) or len(layer.activations) == 0) and hasattr(layer, 'max_batch_len'):
+    if (
+        not isinstance(layer.activations, list) or len(layer.activations) == 0
+    ) and hasattr(layer, "max_batch_len"):
         del layer.max_batch_len
 
 
 def _get_batch_size(layer: nn.Module, grad_sample: torch.Tensor, batch_dim: int) -> int:
     r"""
     Computes and returns the maximum batch size which is the maximum of the dimension values
-    along 'batch_dim' axis over layer.activations + [grad_sample], where layer.activations is 
-    a list. If layer.activations is a not a list, then return grad_sample.shape[batch_dim]. 
+    along 'batch_dim' axis over layer.activations + [grad_sample], where layer.activations is
+    a list. If layer.activations is a not a list, then return grad_sample.shape[batch_dim].
     """
-    
+
     max_batch_len = 0
     if isinstance(layer.activations, list):
         for out in layer.activations:
@@ -261,4 +263,3 @@ def _get_batch_size(layer: nn.Module, grad_sample: torch.Tensor, batch_dim: int)
 
     max_batch_len = max(max_batch_len, grad_sample.shape[batch_dim])
     return max_batch_len
-
--- a/opacus/autograd_grad_sample.py
+++ b/opacus/autograd_grad_sample.py
@@ -23,8 +23,6 @@ from opacus.layers.dp_lstm import LSTMLinear
 from .supported_layers_grad_samplers import _supported_layers_grad_samplers
 from .utils.module_inspection import get_layer_type, requires_grad
 
-from inspect import signature
-
 # work-around for https://github.com/pytorch/pytorch/issues/25723
 _hooks_disabled: bool = False
 
@@ -239,25 +237,5 @@ def _compute_grad_sample(
         get_layer_type(layer)
     )
 
-    if 'max_batch_len' in signature(compute_layer_grad_sample).parameters: # For compatibility with PackedSequences
-        max_batch_len = _get_batch_size(layer, A, batch_dim)
-        compute_layer_grad_sample(layer, A, B, max_batch_len=max_batch_len)
-    else:
-        compute_layer_grad_sample(layer, A, B)        
+    compute_layer_grad_sample(layer, A, B)        
 
-
-def _get_batch_size(layer, A, batch_dim):
-    if batch_dim != 0:
-        return 0
-
-    if hasattr(layer, 'max_batch_len'):
-        return layer.max_batch_len
-
-    max_batch_len = 0
-    for out in layer.activations:
-        if out.shape[0] > max_batch_len:
-            max_batch_len = out.shape[0]
-
-    max_batch_len = max(max_batch_len, A.shape[batch_dim])
-    layer.max_batch_len = max_batch_len
-    return max_batch_len
--- a/opacus/autograd_grad_sample.py
+++ b/opacus/autograd_grad_sample.py
@@ -23,6 +23,7 @@ from opacus.layers.dp_lstm import LSTMLinear
 from .supported_layers_grad_samplers import _supported_layers_grad_samplers
 from .utils.module_inspection import get_layer_type, requires_grad
 
+
 # work-around for https://github.com/pytorch/pytorch/issues/25723
 _hooks_disabled: bool = False
 
--- a/opacus/autograd_grad_sample.py
+++ b/opacus/autograd_grad_sample.py
@@ -23,6 +23,7 @@ from opacus.layers.dp_lstm import LSTMLinear
 from .supported_layers_grad_samplers import _supported_layers_grad_samplers
 from .utils.module_inspection import get_layer_type, requires_grad
 
+from inspect import signature
 
 # work-around for https://github.com/pytorch/pytorch/issues/25723
 _hooks_disabled: bool = False
@@ -214,7 +215,6 @@ def _compute_grad_sample(
 
     # Outside of the LSTM there is "batch_first" but not for the Linear inside the LSTM
     batch_dim = 0 if batch_first or type(layer) is LSTMLinear else 1
-
     if isinstance(layer.activations, list):
         A = layer.activations.pop()
     else:
@@ -238,4 +238,26 @@ def _compute_grad_sample(
     compute_layer_grad_sample = _supported_layers_grad_samplers.get(
         get_layer_type(layer)
     )
-    compute_layer_grad_sample(layer, A, B)
+
+    if 'max_batch_len' in signature(compute_layer_grad_sample).parameters: # For compatibility with PackedSequences
+        max_batch_len = _get_batch_size(layer, A, batch_dim)
+        compute_layer_grad_sample(layer, A, B, max_batch_len=max_batch_len)
+    else:
+        compute_layer_grad_sample(layer, A, B)        
+
+
+def _get_batch_size(layer, A, batch_dim):
+    if batch_dim != 0:
+        return 0
+
+    if hasattr(layer, 'max_batch_len'):
+        return layer.max_batch_len
+
+    max_batch_len = 0
+    for out in layer.activations:
+        if out.shape[0] > max_batch_len:
+            max_batch_len = out.shape[0]
+
+    max_batch_len = max(max_batch_len, A.shape[batch_dim])
+    layer.max_batch_len = max_batch_len
+    return max_batch_len
--- a/opacus/autograd_grad_sample.py
+++ b/opacus/autograd_grad_sample.py
@@ -18,6 +18,7 @@ from typing import Tuple
 
 import torch
 import torch.nn as nn
+
 from opacus.layers.dp_lstm import LSTMLinear
 
 from .supported_layers_grad_samplers import _supported_layers_grad_samplers
--- a/opacus/layers/dp_lstm.py
+++ b/opacus/layers/dp_lstm.py
@@ -159,16 +159,16 @@ class DPLSTMCell(nn.Module):
         )
         i_t = torch.sigmoid(
             i_t_input
-        )  # [B, D] or [batch_size_t, 4*D] if batch_size_t is not None
+        )  # [B, D] or [batch_size_t, D] if batch_size_t is not None
         f_t = torch.sigmoid(
             f_t_input
-        )  # [B, D] or [batch_size_t, 4*D] if batch_size_t is not None
+        )  # [B, D] or [batch_size_t, D] if batch_size_t is not None
         g_t = torch.tanh(
             g_t_input
-        )  # [B, D] or [batch_size_t, 4*D] if batch_size_t is not None
+        )  # [B, D] or [batch_size_t, D] if batch_size_t is not None
         o_t = torch.sigmoid(
             o_t_input
-        )  # [B, D] or [batch_size_t, 4*D] if batch_size_t is not None
+        )  # [B, D] or [batch_size_t, D] if batch_size_t is not None
         if batch_size_t is None:
             c_t = f_t * c_prev + i_t * g_t
         else:
--- a/opacus/layers/dp_lstm.py
+++ b/opacus/layers/dp_lstm.py
@@ -1,10 +1,52 @@
 import math
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, Union
 
 import torch
 import torch.nn as nn
+from torch.nn.utils.rnn import PackedSequence
 
 from .param_rename import ParamRenamedModule
+
+from torch.nn.utils.rnn import pad_sequence, pack_padded_sequence
+
+def compute_seq_lengths(batch_sizes: torch.Tensor):
+    max_batch_size = batch_sizes[0]
+    if len(batch_sizes) == 1:
+        return [1] * max_batch_size
+
+    running_seq = 0
+    running_seq_lengths = []
+    for i in range(1, len(batch_sizes)):
+        delta = batch_sizes[i-1].item() - batch_sizes[i].item()
+        running_seq += 1
+        running_seq_lengths += (delta * [running_seq])
+
+    running_seq += 1
+    running_seq_lengths += (batch_sizes[-1].item() * [running_seq])
+    running_seq_lengths.reverse()
+    return running_seq_lengths
+
+def compute_last_states(h_n: torch.Tensor, c_n: torch.Tensor, seq_lengths: List):
+    max_batch_size = len(seq_lengths)
+    hidden_size = h_n[0].shape[-1]
+    h_last = torch.zeros(max_batch_size, hidden_size)
+    c_last = torch.zeros(max_batch_size, hidden_size)
+
+    for i, seq_len in enumerate(seq_lengths):
+        h_last[i, :] = h_n[seq_len-1][i, :]
+        c_last[i, :] = c_n[seq_len-1][i, :]
+
+    return h_last, c_last
+
+
+def concat_tensor_sequence(a: Union[List, Tuple], b: Union[List, Tuple], dim: int):
+    seq_length = len(a)
+    output = [0] * seq_length
+
+    for i in range(seq_length):
+        output[i] = torch.cat((a[i], b[i]), dim=dim)
+
+    return output
 
 
 class LSTMLinear(nn.Linear):
@@ -49,9 +91,14 @@ class DPLSTMCell(nn.Module):
             nn.init.uniform_(weight, -stdv, stdv)
 
     def forward(
-        self, x: torch.Tensor, h_prev: torch.Tensor, c_prev: torch.Tensor
-    ) -> Tuple[torch.Tensor, torch.Tensor]:
-        gates = self.ih(x) + self.hh(h_prev)  # [B, 4*D]
+        self, x: torch.Tensor, h_prev: torch.Tensor, c_prev: torch.Tensor,
+        B: Optional[int] = None, reverse: Optional[bool] = False) -> Tuple[torch.Tensor, torch.Tensor]:
+
+        if B is None:
+            gates = self.ih(x) + self.hh(h_prev)  # [B, 4*D]
+        else:
+            gates = self.ih(x) + self.hh(h_prev[:B, :])  # [B, 4*D]
+
         i_t_input, f_t_input, g_t_input, o_t_input = torch.split(
             gates, self.hidden_size, 1
         )
@@ -59,7 +106,11 @@ class DPLSTMCell(nn.Module):
         f_t = torch.sigmoid(f_t_input)  # [B, D]
         g_t = torch.tanh(g_t_input)  # [B, D]
         o_t = torch.sigmoid(o_t_input)  # [B, D]
-        c_t = f_t * c_prev + i_t * g_t
+        if B is None:
+            c_t = f_t * c_prev + i_t * g_t
+        else:
+            c_t = f_t * c_prev[:B, :] + i_t * g_t       
+
         h_t = o_t * torch.tanh(c_t)
 
         return h_t, c_t
@@ -92,10 +143,12 @@ class DPLSTMLayer(nn.Module):
 
         self.dropout_layer = nn.Dropout(dropout) if dropout > 0 else None
 
+
     def forward(
         self,
-        x: torch.Tensor,
+        x: Union[torch.Tensor, Tuple],
         state_init: Tuple[torch.Tensor, torch.Tensor],
+        batch_sizes: Optional[torch.Tensor] = None,
     ) -> Tuple[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]:
         r"""
         Implements the forward pass of the DPLSTMLayer when a sequence is given in input.
@@ -105,6 +158,8 @@ class DPLSTMLayer(nn.Module):
             state_init: Initial state of the LSTMCell as a tuple ``(h_0, c_0)``
                 where ``h_0`` is the initial hidden state and ``c_0`` is the
                 initial cell state of the DPLSTMCell
+            batch_sizes: Contains the batch sizes as stored in PackedSequence
+
 
         Returns:
             ``output, (h_n, c_n)`` where, ``output`` is of shape ``[T, B, H]`` and is a
@@ -114,29 +169,57 @@ class DPLSTMLayer(nn.Module):
             tensor containing the cell state for ``t = T``.
         """
 
-        seq_length, batch_sz, _ = x.shape
-        if self.reverse:
-            x = x.flip(0)
-        x = torch.unbind(x, dim=0)
+        if batch_sizes is not None:
+            seq_length = batch_sizes.size(0)
+            if self.reverse:
+                x = tuple(reversed(x))
+                batch_sizes = batch_sizes.flip(0)
+        else:
+            seq_length, batch_sz, _ = x.shape
+            if self.reverse:
+                x = x.flip(0)
+            x = torch.unbind(x, dim=0)
 
         h_0, c_0 = state_init
 
         h_n = [h_0]
         c_n = [c_0]
+        batch_size_prev = h_0.shape[0]
 
         for t in range(seq_length):
-            h_next, c_next = self.cell(x[t], h_n[t], c_n[t])
+            if batch_sizes is not None:
+                batch_size_t = batch_sizes[t].item()
+                delta = batch_size_t - batch_size_prev
+                if delta > 0:
+                    h_cat = torch.cat((h_n[t], h_0[batch_size_prev:batch_size_t, :]), 0)
+                    c_cat = torch.cat((c_n[t], c_0[batch_size_prev:batch_size_t, :]), 0)
+                    h_next, c_next = self.cell(x[t], h_cat, c_cat, batch_size_t, self.reverse)                    
+                else:
+                    h_next, c_next = self.cell(x[t], h_n[t], c_n[t], batch_size_t, self.reverse)
+            else:
+                h_next, c_next = self.cell(x[t], h_n[t], c_n[t])
             if self.dropout:
                 h_next = self.dropout_layer(h_next)
             h_n.append(h_next)
             c_n.append(c_next)
+            batch_size_prev = h_next.shape[0]
 
-        h_n = torch.stack(h_n[1:], dim=0)  # [T, B, H], init step not part of output
+        if batch_sizes is None:
+            h_n = torch.stack(h_n[1:], dim=0)  # [T, B, H], init step not part of output
 
-        return (
-            h_n.flip(0) if self.reverse else h_n,  # Flip the output...
-            (h_n[-1], c_n[-1]),  # ... But not the states
-        )
+            return (
+                h_n.flip(0) if self.reverse else h_n,  # Flip the output...
+                (h_n[-1], c_n[-1]),  # ... But not the states
+            )
+        else:
+            seq_lengths = compute_seq_lengths(batch_sizes)
+            h_temp, c_temp = h_n[1:], c_n[1:]
+            h_last, c_last = compute_last_states(h_temp, c_temp, seq_lengths)
+            if self.reverse:
+                h_temp = tuple(reversed(h_temp))
+            
+            return h_temp, (h_last, c_last)
+
 
 
 class BidirectionalDPLSTMLayer(nn.Module):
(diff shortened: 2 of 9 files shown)


opacus/layers/dp_lstm.py
```diff
@@ -1,11 +1,55 @@
...
+    return running_seq_lengths
+
+
+def compute_last_states(h_n: torch.Tensor, c_n: torch.Tensor, seq_lengths: List):
+    max_batch_size = len(seq_lengths)
+    hidden_size = h_n[0].shape[-1]
+    h_last = torch.zeros(max_batch_size, hidden_size)
+    c_last = torch.zeros(max_batch_size, hidden_size)
+
+    for i, seq_len in enumerate(seq_lengths):
```
Darktex, line 37: This tensor creation via loop is very slow. A way to avoid it is to construct the indices first (you can use a loop for the indices), and then assign them all in one shot

opacus/layers/dp_lstm.py
```diff
@@ -1,11 +1,55 @@
...
+        h_last[i, :] = h_n[seq_len - 1][i, :]
+        c_last[i, :] = c_n[seq_len - 1][i, :]
+
+    return h_last, c_last
+
+
+def concat_tensor_sequence(a: Union[List, Tuple], b: Union[List, Tuple], dim: int):
+    seq_length = len(a)
```
Darktex, line 45: This means that if b is longer than a, we truncate b. Is this what we want?

opacus/layers/dp_lstm.py
```diff
@@ -49,17 +93,31 @@ def reset_parameters(self):
             nn.init.uniform_(weight, -stdv, stdv)
 
     def forward(
-        self, x: torch.Tensor, h_prev: torch.Tensor, c_prev: torch.Tensor
+        self,
+        x: torch.Tensor,
+        h_prev: torch.Tensor,
+        c_prev: torch.Tensor,
+        B: Optional[int] = None,
```
Darktex, line 100: Who's B? What does it do?

opacus/tests/dp_layers_test.py
```diff
@@ -253,6 +448,81 @@ def _reset_seeds(self):
         torch.manual_seed(1337)
         torch.cuda.manual_seed(1337)
 
+    def _test_PackedSequence(self, seq1, seq2, is_sorted=True):
```
Darktex, line 451: Great test!

opacus/layers/dp_lstm.py
```diff
@@ -94,8 +152,9 @@ def __init__(
 
     def forward(
         self,
-        x: torch.Tensor,
+        x: Union[torch.Tensor, Tuple],
```
Darktex, line 155: One more thing: why is `x` a Tuple and not a PackedSequence directly?

opacus/layers/dp_lstm.py
```diff
@@ -114,29 +175,60 @@ def forward(
...
+            seq_lengths = compute_seq_lengths(batch_sizes)
+            h_temp, c_temp = h_n[1:], c_n[1:]
+            h_last, c_last = compute_last_states(h_temp, c_temp, seq_lengths)
+            if self.reverse:
+                h_temp = tuple(reversed(h_temp))
+
+            return h_temp, (h_last, c_last)
 
 
 class BidirectionalDPLSTMLayer(nn.Module):
```
Darktex, line 234: In my opinion, this is the first class where we can have an `if` and switch between the "normal" layers or the ones for PackedSequences

opacus/layers/dp_lstm.py
```diff
@@ -94,8 +152,9 @@ def __init__(
 
     def forward(
         self,
-        x: torch.Tensor,
+        x: Union[torch.Tensor, Tuple],
         state_init: Tuple[torch.Tensor, torch.Tensor],
+        batch_sizes: Optional[torch.Tensor] = None,
```
Darktex, line 157: Couldn't we just read them from x? If x is a PackedSequence, then these will be waiting for us in `x.batch_sizes`: https://pytorch.org/docs/stable/generated/torch.nn.utils.rnn.PackedSequence.html

opacus/layers/dp_lstm.py
```diff
@@ -182,6 +274,7 @@ def forward(
         self,
         x: torch.Tensor,
         state_init: Tuple[torch.Tensor, torch.Tensor],
+        batch_sizes: Optional[torch.Tensor] = None,
```
Darktex, line 277: Same as above: if x is packed, we have this already

opacus/layers/dp_lstm.py
```diff
@@ -349,8 +463,29 @@ def forward(
...
+            self._permute_hidden(hs, unsorted_indices),
+            self._permute_hidden(cs, unsorted_indices),
+        )
+
+    def _permute_hidden(
+        self, x: torch.Tensor, permutation: Optional[torch.Tensor] = None, dim: int = 1
+    ):
+        if permutation is None:
+            return x
+        return x.index_select(dim, permutation)
```
Darktex, line 488: Using this function is not a good idea. It makes a copy of the tensor! https://pytorch.org/docs/stable/generated/torch.index_select.html

opacus/layers/dp_lstm.py
```diff
@@ -320,6 +431,8 @@ def forward(
             )
         else:
             h_0s = h_0s.reshape([L, P, B, H])
+            h_0s = self._permute_hidden(h_0s, sorted_indices, 2)
```
Darktex, line 434: Why do we need this?

opacus/layers/dp_lstm.py
```diff
@@ -301,8 +400,20 @@ def forward(
             - ``c_n`` is of shape ``[L * P, B, H]`` and contains the cell state for ``t = T``.
         """
 
-        x = self._rearrange_batch_dim(x)
-        T, B, D = x.shape
+        if isinstance(x, PackedSequence):
+            x, batch_sizes, sorted_indices, unsorted_indices = x
```
Darktex, line 404: No reason to "unpack" the sequence here I think. Just send it down as-is, and let the actual class that needs to work with it handle it

opacus/layers/dp_lstm.py
```diff
@@ -114,29 +175,60 @@ def forward(
...
 
         h_0, c_0 = state_init
 
         h_n = [h_0]
         c_n = [c_0]
+        batch_size_prev = h_0.shape[0]
 
         for t in range(seq_length):
-            h_next, c_next = self.cell(x[t], h_n[t], c_n[t])
+            if batch_sizes is not None:
```
Darktex, line 196: Let's handle a PackedSequence directly here, and let's just put this code into a helper function so that this stays readable

Darktex (changes requested):
Thank you for your contribution @touqir14 ! We all really appreciate it :) 

I have a few comments mostly to improve code style and readability and then we are good to merge!

opacus/layers/dp_lstm.py
touqir14: Actually, the type hints for ```h_n``` and ```c_n``` in ```compute_last_states``` are typos and they are both a list of tensors. I will fix that in my next commit. I think you suggested the above optimization thinking they are tensors. Being a list of tensors of varying shapes, I don't think we can apply the optimization you suggested. Is there a torch function for doing that? I couldn't find one.

opacus/layers/dp_lstm.py
```diff
@@ -1,11 +1,55 @@
...
 import torch
 import torch.nn as nn
+from torch.nn.utils.rnn import PackedSequence
 
 from .param_rename import ParamRenamedModule
 
+from torch.nn.utils.rnn import pad_sequence, pack_padded_sequence
+
+
+def compute_seq_lengths(batch_sizes: torch.Tensor):
```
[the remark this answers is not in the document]
touqir14: Sure, I will add them in the next commit.

opacus/layers/dp_lstm.py
touqir14: ```concat_tensor_sequence``` is being used in the ```BidirectionalDPLSTMLayer``` class and it takes the output sequences from the forward and reverse layer and so the lengths of two output sequences are the same. So, we are good for now. In case that we want to use this function for any other case, would you like me to throw some warning or exception in case ```a``` and ```b``` are of different lengths?

opacus/layers/dp_lstm.py
touqir14: Since ```B``` was used to denote the batch size, I used it here. This variable represents the batch size at time step t. I will rename it to ```batch_size_t```. When PackedSequence is not passed into the DPLSTM, this variable will be ```None``` and so the ```forward``` function will operate like before. Since the rows of ```h_prev``` and ```c_prev``` are sorted based on the sequence lengths, ```B``` is used to select the values in h_prev and c_prev corresponding to only the top ```B``` sequences.

touqir14:
@Darktex , So I have ensured that only ```DPLSTM``` gets to see and handle(encode, decode) the PackedSequence and as such it passes down a tuple of x_i (for i = 1 to max sequence length), where x_i is of shape (batch size at timestep i, dimension of the encoding) to the DPLSTMLayer or the bidirectional version. So, the layers themselves just operate on these tuples. One advantage of this is that if we have multiple layers, we avoid encoding and decoding PackedSequence multiple times, which is better for performance. The decoding part is given in the ```DPLSTM.forward``` function : 
```python
if isinstance(x, PackedSequence):
    x, batch_sizes, sorted_indices, unsorted_indices = x
    B = batch_sizes[0].item()
    seq_length = batch_sizes.size(0)
    batch_sz = None
    _, D = x.shape
    x = x.split(tuple(batch_sizes))
```
And the encoding part within ```DPLSTM.forward```
```python
if batch_sizes is not None:
    seq_lengths = compute_seq_lengths(batch_sizes)
    packed_data = pack_padded_sequence(
        pad_sequence(x, batch_first=False), seq_lengths, batch_first=True
    )[0]
    out = PackedSequence(
        packed_data, batch_sizes, sorted_indices, unsorted_indices
    )
```
Secondly, if we instead carry out the encoding and decoding part within each layer, things get complicated for reverse layers due to the formatting of the data within PackedSequence and I am not yet sure how to decode properly taking the reverse direction into account. I considered these two factors before deciding on the current approach. I also think that this is a more elegant approach since the Layers have lesser responsibility which should also make it potentially easier to extend or create new Layer classes without having to worry about PackedSequence data formats. Let me know what you think.

opacus/layers/dp_lstm.py
touqir14: So, the PackedSequence always ensures that each of the "batches" within the 2 dimensional data tensor contain the encoding of the sequence in descending order of the sequence lengths even though the original data did not contain sequences in a sorted order(descending) of sequence length. Let me clarify with an example:
```python
from opacus.layers import DPLSTM
import torch 

seq_batch = [torch.tensor([[1, 1],
                           [2, 2],
                           [3, 3],
                           [4, 4],
                           [5, 5]]),
             torch.tensor([[10, 10],
                           [20, 20]])]

seq_lens = [5, 2]

padded_seq_batch = torch.nn.utils.rnn.pad_sequence(seq_batch, batch_first=True)
packed_seq_batch = torch.nn.utils.rnn.pack_padded_sequence(padded_seq_batch, lengths=seq_lens, batch_first=True)


seq_batch = [torch.tensor([[10, 10],
                           [20, 20]]),
            torch.tensor([[1, 1],
                           [2, 2],
                           [3, 3],
                           [4, 4],
                           [5, 5]])]

seq_lens = [2, 5]

padded_seq_batch = torch.nn.utils.rnn.pad_sequence(seq_batch, batch_first=True)
packed_seq_batch = torch.nn.utils.rnn.pack_padded_sequence(padded_seq_batch, lengths=seq_lens, batch_first=True, enforce_sorted=False)

```
Both approaches produce the same tensor within the PackedSequence. The ```_permute_hidden``` function makes sure that the arrangement of the hidden weights are in line with the data format within PackedSequence.

opacus/layers/dp_lstm.py
touqir14: Good catch, I will fix that.

opacus/layers/dp_lstm.py
Darktex: Ah, that makes sense. How about renaming it to `_concat_sequence_directions`? We could also call `a` and `b` something like `forward` and `reverse` so that this is more explicit. At that point, the `if` check that they are of the same length would immediately make sense to the reader

touqir14: In addition to the above, I have changed the other two function names to have an underscore at the beginning in my recent commit.

opacus/layers/dp_lstm.py
Darktex: Yeah, not much to do then :) Should be fine

opacus/layers/dp_lstm.py
```diff
@@ -1,11 +1,102 @@
...
+    c_last = torch.zeros(max_batch_size, hidden_size)
+
+    for i, seq_len in enumerate(seq_lengths):
+        h_last[i, :] = h_n[seq_len - 1][i, :]
+        c_last[i, :] = c_n[seq_len - 1][i, :]
+
+    return h_last, c_last
+
+
+def _concat_sequence_directions(
```
Darktex, line 71: (y)

opacus/tests/dp_layers_test.py
```diff
@@ -103,6 +104,81 @@ def _reset_seeds(self):
...
+        for seq1_param, seq2_param, message in seq_to_test:
+            assert_allclose(
+                actual=seq1_param,
+                expected=seq2_param,
+                atol=10e-6,
+                rtol=10e-5,
+                msg=f"Tensor value mismatch between {message}",
+            )
+
+    def _gen_packedData(self, minibatch_size, max_seq_length, input_dim, sorted_=False):
```
Darktex, line 153: nit: rename to `_gen_packed_data` (ie let's use lower_snake_case)

opacus/tests/dp_layers_test.py
```diff
@@ -8,6 +8,7 @@
 from torch.nn import LSTM
 from torch.nn.modules.activation import MultiheadAttention
 from torch.testing import assert_allclose
+from torch.nn.utils.rnn import pad_sequence, pack_padded_sequence
```
Darktex, line 11: After landing #120  we have rewritten these tests to use Hypothesis. Can we rebase this on top of Master and update these tests?

I would do it by making a new bool flag `packed_input`, and calling `_gen_packed_data` on the input and `pad_packed_sequence` on the output if it's True (I think we are going to need to latter to use `torch.testing.all_close`? I doubt it's smart enough to accept PackedSequences directly). Everything else should be the same I think? 

Note: you can have another flag for "sorted_sequence", and then in the body of the function just check if `pack_sequence = True`. If not, disregard that flag

opacus/layers/dp_lstm.py
Darktex: Makes sense :) Thanks for explaining!

Darktex (changes requested):
Thank you for the lengthy explanation @touqir14 ! I agree with you that the cost of unpacking per-layer is not worth it, and I think we are almost ready to merge :) 

Only a couple of comments left, the major of which is realigning the tests to the new version since #120 changed those tests. It should be quite straightforward though :)

Note: This week was a bit messy for me but do expect a much faster review next time

opacus/tests/dp_layers_test.py
touqir14: I have updated the tests @Darktex . So, I have instead added a flag "packed_input_flag" which is 0 for standard input, 1 for sorted PackedSequence, and 2 for unsorted PackedSequence in order to keep the configuration space small.
I had to adapt the ```DPModules_test``` class within ```common.py``` for testing PackedSequence data. See the ```DPModules_test._check_packed_sequence``` method.

Darktex (approved):
Approved! Thank you so much for this awesome contribution :)

opacus/tests/dp_layers_test.py
Darktex: Works great!

Darktex:
I was like "why isn't this merged yet?", then I remembered that we are supposed to merge on our side lol.

Sorry D: merging now

ffuuugor (changes requested):
Hey @touqir14 
Thanks again for your amazing contribution - it's great to have this implemented.
However, it seems to me, that this PR needs some more work before we can land it.

I was experimenting with the code in various settings, and it looks to me as if PackedSequence doesn't play well with PrivacyEngine.
Specifically, in my tests DPLSTM:
1) works fine with PrivacyEngine + regular data
2) works fine without PrivacyEngine + packed data
3) crashes with PrivacyEngine + packed data

I think the right way forward is to add a new unit test to check how DPLSTM+packed inputs perform when training with privacy.
There are two possible ways to do that, both equally good IMO.

One is to build toy model test case, in the same manner we do in `privacy_engine_test.py`. This test creates a model with various supported layers (`SampleConvNet`), runs a few iterations of mock training and check various invariants (e.g. per-sample gradients sum up to an original autograd gradient.).
It already abstracts out model and data handing, so it should be pretty easy to extend this test to cover another model with a different data factory.

Another option is to extend `dp_lstm_test.py`. You've already did a great job testing packed sequences there, so the only missing piece is training. You'll need to add an actual optimizer (not only `loss.backward()`) and optionally attach PrivacyEngine (probably with no-op noise and clipping)

touqir14:
@ffuuugor , can you share the code snippet that causes the crash?

ffuuugor:
Let me know if I'm using it wrong

```python
from opacus.layers import DPLSTM
from opacus import PrivacyEngine
import torch 
from torch import nn

seq_batch = [torch.tensor([[1, 1],
                           [2, 2],
                           [3, 3],
                           [4, 4],
                           [5, 5]]),
             torch.tensor([[10, 10],
                           [20, 20]])]

seq_lens = [5, 2]

padded_seq_batch = torch.nn.utils.rnn.pad_sequence(seq_batch, batch_first=True)
packed_seq_batch = torch.nn.utils.rnn.pack_padded_sequence(padded_seq_batch, lengths=seq_lens, batch_first=True)

dp_lstm = DPLSTM(input_size=2, hidden_size=3, batch_first=True)

optimizer = torch.optim.SGD(dp_lstm.parameters(), lr=0.1, momentum=0)
criterion = nn.MSELoss()

privacy_engine = PrivacyEngine(
    dp_lstm,
    batch_size=2,
    sample_size=100,
    alphas = [1,2,3],
    noise_multiplier=1,
    max_grad_norm=1,
    batch_first=True
)

privacy_engine.attach(optimizer)

output, (hn, cn) = dp_lstm(packed_seq_batch.float())
y = torch.zeros_like(output[0])
loss = criterion(output[0], y)
loss.backward()
```

touqir14:
The code runs fine without attaching the privacy engine. With the privacy engine, I am getting this error:
```RuntimeError: output with shape [1, 12, 3] doesn't match the broadcast shape [2, 12, 3]``` . I am guessing this needs some modifications to the privacy engine. I will take a look.

**Comment by touqir14** [comment]
Regarding the crash that you observed @ffuuugor , I think this is a more fundamental issue and is not caused by PackedSequences. See issue [134](https://github.com/pytorch/opacus/issues/134)

**Comment by ffuuugor** [comment]
Huh, that's an interesting issue.
But from what I can see so far, the issue is unrelated to the crash I've had. 

a) #134 affects only python3.9, while my code crashes on 3.6.9 too
b) #134 also affects nn.LSTM, while my code works fine with non-DP version
c) #134 crashes upon env destruction, after the code has been executed. While issue I've described is clearly happening on autograd step

**Comment by touqir14** [comment]
The privacy engine with PackedSequences should be fine now. What was causing the compatibility issue with privacy engine was the per sample gradient computation part which needed some adjustments. I have added the tests to the ```privacy_engine_tests.py``` file. I also changed a few variable names in that file to improve the segregation between the CNN code and LSTM code. However, I haven't yet encountered any segmentation faults or crashes other than #134 and the only error was the runtime error due to the tensor dimension mismatches earlier. The tests were carried out with python 3.7 and torch 1.8 . Let me know if you are still facing any crashes.

**Review by ffuuugor — CHANGES_REQUESTED** [review]
Thanks for fixing this! I can confirm that my use case no longer crashes.
It's also great that you've added extra tests to `privacy_engine_test.py`, it'll help make sure we don't break this inadvertently in the future.

That said, I think the structure of the new tests can be improved. 
You've added new checks to the existing methods, which leads to two problems:
1) As a rule of thumb, we want to keep `test_*` methods as short and concise as possible. We don't want to check for many different things within one method - it makes test results harder to read and debug. 
2) This creates a fair amount of code duplication.

What we ideally want for this test is to abstract out the common parts (e.g. model initialization and training, gradient checks) and only implement model and data factory for each individual test model.
Really good example of this approach is our tests in for grad samples. Here we have a base class (`GradSampleHooks_test`) which takes care of all the common boilerplate code to run the test. We then subclass it, implementing only the necessary bits to tailor to a custom test case.
I believe here it could work in a similar fashion - you can create base class, leaving abstract methods for model and data, and then subclass it for 3 different models: SampleConvNet, unidirectional LSTM and bidirectional LSTM.

Let me know what you think about this

File: opacus/supported_layers_grad_samplers.py [hunk 17]
```diff
@@ -59,18 +58,25 @@ def _create_or_accumulate_grad_sample(
...
+        else:
+            param.grad_sample = torch.zeros(torch.Size([max_batch_len]) + grad_sample.shape[1:])
+            param.grad_sample[:grad_sample.shape[0]] = grad_sample        
     else:
-        param.grad_sample = grad_sample.clone()
+        if hasattr(param, "grad_sample"):
+            param.grad_sample += grad_sample
+        else:
+            print("clonning!")
+            param.grad_sample = grad_sample.clone()
```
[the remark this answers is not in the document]
touqir14: Right, I must have forgotten to remove it.

opacus/autograd_grad_sample.py
```diff
@@ -238,4 +238,26 @@ def _compute_grad_sample(
...
+    if hasattr(layer, 'max_batch_len'):
+        return layer.max_batch_len
+
+    max_batch_len = 0
+    for out in layer.activations:
+        if out.shape[0] > max_batch_len:
+            max_batch_len = out.shape[0]
+
+    max_batch_len = max(max_batch_len, A.shape[batch_dim])
+    layer.max_batch_len = max_batch_len
```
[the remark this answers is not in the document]
touqir14: I am not sure how this can be done for one batch only. For now, I have removed caching. If in the future this part gets object oriented, the batch-level caching should be easily done. Its quadratic in the number of sequence lengths. The overhead should not be noticeable when dealing with large batch sizes and/or large input dimensions (since these parts will dominate when updating ```param.grad_sample```). For example, I just benchmarked the ```_get_batch_size``` function on an activations list of 1 million entries and found that the function took only about 160ms for one call. We rarely see sequence lengths of a million, if at all!

opacus/autograd_grad_sample.py
```diff
@@ -238,4 +238,26 @@ def _compute_grad_sample(
     compute_layer_grad_sample = _supported_layers_grad_samplers.get(
         get_layer_type(layer)
     )
-    compute_layer_grad_sample(layer, A, B)
+
+    if 'max_batch_len' in signature(compute_layer_grad_sample).parameters: # For compatibility with PackedSequences
```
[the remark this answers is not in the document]
touqir14: Updated this.

opacus/autograd_grad_sample.py
```diff
@@ -238,4 +238,26 @@ def _compute_grad_sample(
...
+        max_batch_len = _get_batch_size(layer, A, batch_dim)
+        compute_layer_grad_sample(layer, A, B, max_batch_len=max_batch_len)
+    else:
+        compute_layer_grad_sample(layer, A, B)        
+
+
+def _get_batch_size(layer, A, batch_dim):
+    if batch_dim != 0:
+        return 0
```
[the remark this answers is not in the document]
touqir14: Yeah, I must have overlooked line 234-236.

opacus/tests/privacy_engine_test.py
```diff
@@ -74,46 +147,115 @@ def forward(self, x):
     def name(self):
         return "SampleConvNet"
 
+class SampleLSTMs():
+    input_size = 20
+    max_seq_length = 30
+    def __init__(self, model_initializer, **kwargs):
+        self.criterion = nn.MSELoss()
+        self.init_lstms()
+        fn_kwargs = kwargs.copy()
```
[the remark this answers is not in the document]
touqir14: Its use case is when you are creating a DP version of LSTM. For doing that I am passing a list of ```state_dict``` for all the non-private LSTMs in the kwargs dictionary and since I am individually initializing the private models and their corresponding optimizers, ```fn_kwargs``` only need to have the ```state_dict``` for the corresponding non-private lstm model.

touqir14:
Just pushed two commits addressing the small suggestions. Its good to know that the code is working fine on your end. Coming to the refactoring part of ```privacy_engine_test.py```, I think this deserves a separate PR given the amount of changes required. Also, I didn't see any tests that actually compare the closeness of the weights and their gradients between the private and non-private models for various levels of epsilon and delta. Some guarantees can definitely be laid out that holds with high probability using standard concentration of measure analysis. This would really be awesome to have that can verify the correctness of the privacy engine. All in all, I think a separate PR, which is more dedicated to privacy engine, is the better path to take.

opacus/autograd_grad_sample.py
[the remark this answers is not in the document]
ffuuugor: I was thinking, that caching could work if we reset the cache on `zero_grad()`?
But I agree, that the overhead shouldn't be too bad and we don't have to solve it now

opacus/tests/privacy_engine_test.py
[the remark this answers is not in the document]
ffuuugor: Oh, I see
Thanks for explaining

ffuuugor:
> I think this deserves a separate PR given the amount of changes required.

I'm afraid I'll have to disagree on this. Two things:
1) We do need a test case for computing grad samples with packed sequences. As we've seen, it's non-trivial case and could lead to problems. We need a way to make sure it works correctly
2) Current changes to `privacy_engine_test.py` hurt readability and make it harder to maintain the test.

As I've mentioned before, there are few different ways to test grad_sample + packed sequence. If you feel like `privacy_engine_test.py` approach would be too cumbersome, you can look at adding packed_input option to `tests/grad_samples/dp_lstm_test.py`. (Previously I also pointed at `tests/dp_layers/dp_lstm_test.py`, but I have to admit I've been wrong - grad_sample tests are a better fit for it)

We really appreciate you taking the time to work on this, and I recognize commitment to great code quality in this PR. It's just we believe that high code standards should apply to unit tests too.

touqir14:
I think you are right that we should focus on per gradient sample tests for PackedSequences. In that case we won't have to add lstm tests in the privacy engine test file since the current set of tests in the privacy test file tests the correctness of the privacy engine as a whole and along with per gradient sample tests for PackedSequences should imply that privacy engine will work fine along with DPLSTM using PackedSequences as input, right? If so, adding test cases in ```tests/grad_samples/dp_lstm_test.py``` as you said earlier should be sufficient. PR #125 is migrating the gradient sample test cases to hypothesis. For this PR, shall I simply add test cases without worrying about hypothesis styled ones? ```tests/grad_samples/common.py``` will require adjustments for checking correctness using PackedSequences.

ffuuugor:
Yeah, that sounds good - adding new grad_sample test would suffice. Thank you!
As for #125, either way is fine, the merging should be simple enough. If you don't want to get blocked on landing #125 now, just add non-hypothesis tests and we can merge this PR

touqir14:
@ffuuugor , I am now caching ```max_batch_size``` inside the ```_compute_grad_sample``` function of ```autograd_grad_sample.py``` file in a data dependent way which is good.  The ```tests/grad_samples/common.py``` file needed some adjustments mostly because of the 'mean' reduced L1 loss. Looks like the grad samples test are now working fine with PackedSequences.

ffuuugor (approved):
Looks great! Thanks a lot for working on this and for your patience with the reviews.
Can you pls rebase onto master, so we can land this

touqir14:
Rebasing done!

opacus/tests/dp_layers/dp_lstm_test.py
```diff
@@ -69,11 +69,21 @@ def test_lstm(
 
         dp_lstm.load_state_dict(lstm.state_dict())
 
-        x = (
-            torch.randn([batch_size, seq_len, emb_size])
-            if batch_first
-            else torch.randn([seq_len, batch_size, emb_size])
-        )
+        if packed_input_flag == 0:
```
ffuuugor, line 72: It seems like rebase didn't go well - `packed_input_flag` and `_gen_packed_data ` are missing.

Also, given that `_gen_packed_data` is now used in two different tests, it might be a good idea to move it to utils

opacus/layers/dp_lstm.py
```diff
@@ -303,8 +451,20 @@ def forward(
             the cell state for ``t = T``.
         """
 
-        x = self._rearrange_batch_dim(x)
-        T, B, D = x.shape
+        if isinstance(x, PackedSequence):
+            x, batch_sizes, sorted_indices, unsorted_indices = x
+            B = batch_sizes[0].item()
+            seq_length = batch_sizes.size(0)
+            batch_sz = None
```
ffuuugor, line 458: It looks like these 2 variables are not used - linter says hi

opacus/tests/grad_samples/dp_lstm_test.py
```diff
@@ -6,6 +6,79 @@
...
+                    and 'unsorted_indices' fields will be None.
+
+    Return Value:
+        packed_data : A PackedSequence object with its data sampled from a normal distribution.
+    """
+
+    if batch_first:
+        data = []
+        seq_lengths = []
+        for i in range(minibatch_size):
```
ffuuugor, line 38: linter will complain about `i` not being used within the loop. Please use `for _ in range(minibatch_size)` instead

opacus/tests/grad_samples/dp_lstm_test.py
```diff
@@ -6,6 +6,79 @@
...
+            seq_length = torch.randint(1, max_seq_length + 1, (1,)).item()
+            seq_lengths.append(seq_length)
+            data.append(torch.randn(seq_length, input_dim))
+
+        if sorted_:
+            key = lambda x: x.shape[0]
+            data = sorted(data, key=key, reverse=True)
```
ffuuugor, line 45:
```suggestion
            data = sorted(data, key=lambda x: x.shape[0], reverse=True)
```

opacus/layers/dp_lstm.py
```diff
@@ -1,11 +1,102 @@
...
+    reverse: Union[List[torch.Tensor], Tuple[torch.Tensor]],
+    dim: int,
+) -> Tuple[torch.Tensor]:
+    r"""
+    Given two list/tuple of same length containing tensors, this function returns a concatenation along dimension d. So, output[i] : concatenation of forward[i] and reverse[i] along dimension dim.
+    forward[i] and reverse[i] should have the same shape. This function is used for concatenating the outputs of the forward and reverse layer of a bidirectional LSTM.
+
+    Args:
+        forward: list/tuple containing n tensors, representing the output of the forward layer.
+        reverse: list/tuple containing n tensors, representing the output of the backward layer.
```
ffuuugor, line 82: `dim` is missing from args list

ffuuugor (changes requested):
Hey. Sorry, have to request changes once more - the rebase might've broken the tests

touqir14:
I pushed an update. Let me know if it fixes the issues.

ffuuugor:
Github says the branch is out of date with master again, can you pls rebase?

touqir14:
rebased it.

ffuuugor (approved):

ffuuugor:
@touqir14 Unfortunately, we had to unland these changes (see #141), as they cause issues with one of our integration tests (https://app.circleci.com/pipelines/github/pytorch/opacus/602/workflows/b196ef1d-78dc-4969-8c82-235d54af43fb/jobs/2848)

touqir14:
Hmm, that was unexpected. I am not sure the root cause though. I can check this out later this week or the next. I think @Darktex mentioned #141 that this issue occurs when run only on GPU.

ffuuugor:
Thanks! We'd really appreciate if you can help us drive this to the finish line

The failing pattern in interesting - it seems to be failing on both CPU and GPU, but for different reasons.
On [GPU](https://app.circleci.com/pipelines/github/pytorch/opacus/612/workflows/85e5161b-24eb-4b06-8b95-feb1344a2b12/jobs/2899) you get device mismatch error:
`Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!`

On [CPU](https://app.circleci.com/pipelines/github/pytorch/opacus/602/workflows/b196ef1d-78dc-4969-8c82-235d54af43fb/jobs/2848) you get shape mismatch error:
 `
stack expects each tensor to be equal size, but got [759] at entry 0 and [1600] at entry 1`

touqir14:
So the charlstm example in the CI shows the CPU error. Could you please refer me to the GPU error message?

ffuuugor:
https://app.circleci.com/pipelines/github/pytorch/opacus/612/workflows/85e5161b-24eb-4b06-8b95-feb1344a2b12/jobs/2899

touqir14:
Thanks. I mistakenly deleted the prior message where i asked for the GPU error message link.

touqir14:
@ffuuugor , the fix for the GPU bug was just changing 1 line. After that fix, both the GPU and CPU instances are reproducing the same error. After doing some debugging, I am noticing that after some point during training phase, the ```layer.activations``` inside ```autograd_grad_sample._compute_grad_sample``` keeps containing residual tensors. For instance, after some point ```layer.activations = [x_0, ..., x_i, y_k,...]``` will partially remain unchanged : ```x_0 , ... x_i``` won't change whereas ```y_k,...``` tensors will change . I was under the impression that the gradient samples will be popped out from ```layer.activations``` one by one after each consecutive call to ```_compute_grad_sample``` so that once ```layer.activations``` becomes an empty list, the model will be run on the minibatch to fill up ```layer.activations```.

What is the purpose of keeping residual tensors in ```layer.activations```? Maybe I am misunderstanding something. The prior fixes for ```_compute_grad_sample``` in order to ensure it works properly with PackedSequences assumed that no residual tensors will be left behind.

If you print out the shapes of the tensors present within ```layer.activations``` every time ```grad_sample``` is cloned as shown below, you will notice that after some point during training, ```layer.activations``` keeps storing the residual tensors : https://github.com/pytorch/opacus/blob/master/opacus/supported_layers_grad_samplers.py#L48-L65

ffuuugor:
Thank you for the investigation!
I think you're right, that this is not an intended behaviour. It looks like the problem with the test itself, not with your code.
I've fixed those in #146 - let's wait until it's landed and see if this PR passes all the checks. 
In the meantime, I suggest you commit GPU-specific fix - all goes well it's the only thing needs fixing

touqir14:
pushed the GPU fix and rebased.

touqir14:
The GPU error is gone. There is a linting error for which I will push a fix later (in case more fixes are needed for CPU).

touqir14:
@ffuuugor , would be great to see the "requests" import error fixed for the MNIST example. Also, I am not sure why isort is raising errors even after I fixed them with isort.

ffuuugor:
Yeah, we're working on it.
In the meantime, can you pls fix linter issues: https://app.circleci.com/pipelines/github/pytorch/opacus/629/workflows/56771560-7eda-4cdd-9a00-4ca56e7e116e/jobs/3009

touqir14:
closing/reopening to restart CI

touqir14:
There is something fishy with the lint checks in CI. Even after running isort, the CI is giving me isort errors with ```autograd_grad_sample.py``` . The import ordering of the master and my last commit is the same : 
Master : https://github.com/pytorch/opacus/blob/master/opacus/autograd_grad_sample.py#L16-L24

PR's last commit : https://github.com/pytorch/opacus/blob/a3bf599d8c08ae79d4ea2f582ac4f2a2b5e94fe5/opacus/autograd_grad_sample.py#L16-L25

ffuuugor:
isort checks not only the ordering, but line separation as well.
I believe the issue here is the extra line you've added. I think the idea is to have three import groups (built-in packages, 3rd party packages and dot-imports) each separated by a new line

ffuuugor:
btw. you can try rebasing the branch - MNIST test issue should be fixed now

touqir14:
Closing/Reopening to restart CI

touqir14:
@ffuuugor , looks good from here. You can import the PR for further testing.

ffuuugor:
It does look good indeed. Thanks a lot!

One very last ask. Can you please re-create this PR as a brand new pull request? Our infra doesn't support landing the same PR twice

PR #128, opened 2021-02-02
Fix batch construction and privacy engine

ConstanceBeguier:
## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

In this PR, we fixed the arguments of the privacy engine in order
to be able to use the correct batch construction. We also update
the batch construction used in all examples and tutorials.

The privacy budget evaluation implemented in Opacus follows [1].
This evaluation is based on the Sampled Gaussian Mechanism. As a
consequence, this analysis assumes that to build one batch, each
sample in the dataset is selected with probability q. In addition,
each batch is built independently from the other batches. Using
another batch construction will break this assumption and the
privacy budget will then be incorrectly evaluated. In particular,
as pointed out in [2], one cannot sample constant-sized batches
through random dataset shuffling at each epoch and use the privacy
accounting method from [1].

[1] Rényi Differential Privacy of the Sampled Gaussian Mechanism
[2] Improving deep learning with differential privacy using gradient
encoding and denoising

This commit has been tested on examples and tutorials.

## Checklist

All tests except one passed. This failed test was already broken before this commit.


opacus/utils/uniform_sampler.py
```diff
@@ -0,0 +1,34 @@
+import torch
+from torch.utils.data import Sampler
+
+
+class UniformWithReplacementSampler(Sampler):
```
Darktex, line 5: How is this different from the standard `RandomSampler` in PyTorch? https://pytorch.org/docs/stable/data.html#torch.utils.data.RandomSampler

It can accept `replacement=True` (though normally people use it with that set to False)

opacus/privacy_engine.py
```diff
@@ -31,15 +31,14 @@ class PrivacyEngine:
...
+        >>> privacy_engine = PrivacyEngine(model, sample_rate, alphas=range(2,32), noise_multiplier=1.3, max_grad_norm=1.0)
         >>> privacy_engine.attach(optimizer)  # That's it! Now it's business as usual.
     """
 
     def __init__(
         self,
         module: nn.Module,
-        batch_size: int,
-        sample_size: int,
+        sample_rate: float,
```
Darktex, line 41: Haha this is a point of contention indeed, and something that we kept going back to. If we follow the Abadi et al convention, then yes we should use `sample_rate` in lieu of `batch_size` and `len(dataset)`. On the other hand, PyTorch really treats the latter convention as a first-class citizen, and it shows everywhere (for example, you design models thinking about your batch dimension, not a sample rate). So the decision was to make the API more PyTorch-friendly, given that the sample rate can easily be computed from those two anyway.

Curious to hear your perspective on this

examples/dcgan.py
```diff
@@ -162,12 +169,16 @@
...
 else:
     generator = None
+
 dataloader = torch.utils.data.DataLoader(
     dataset,
-    batch_size=opt.batch_size,
-    shuffle=True,
     num_workers=int(opt.workers),
     generator=generator,
+    batch_sampler=UniformWithReplacementSampler(
```
Darktex, line 177: Yeah, either with this sampler or with the default PyTorch sampler (see comment below), the sticking point is that currently we would leave configuring dataloading to the client, without Opacus helping you in any way.

This is fine and not something we will handle in this review, but we have been thinking for a while to have Opacus help on this too, either by attaching to a dataloader, or by accepting a dataloader in input and returning a new dataloader with all these settings fixed for you, etc.

We could really use extra input on this :)

Darktex (commented):
Thank you for the contribution @ConstanceBeguier! It's really well put together and thanks for updating all examples and even the notebooks! :) 

I have some questions pertaining the API and some details on the sampler, but overall this looks very good to me

opacus/utils/uniform_sampler.py
ConstanceBeguier: I did not succeed to implement the batch construction from Mironov et al [1] with RandomSampler. With this convention, each batch is sampled according to the Sampled Gaussian Mechanism SGM. That means that to build one batch, each sample in the dataset is selected with probability q (called sample rate). Thus, the size of each batch is not fixed. In addition, each batch is built independently from the other batches.

If you are able to implement batch construction according to SGM with RandomSampler, I am very interesting to have access to this implementation.

[1] Rényi Differential Privacy of the Sampled Gaussian Mechanism

opacus/privacy_engine.py
ConstanceBeguier: It is  exactly what I would like to fix with this PR. The main bug in Opacus library is the fact that the batch construction from Abadi et al [2] is used with the privacy budget evaluation from Mironov et al [1]. This privacy budget evaluation [1] is based on the Sampled Gaussian Mechanism. As a consequence, this evaluation assumes that to build one batch, each sample in the dataset is selected with probability q (called sample rate). With this batch construction, the batch size does not exist because the batches will have variable size. Using
another batch construction will break this assumption and the privacy budget evaluation based on Mironov et al [1] will be incorrect (as pointed out by Nasr et al in [3]).

To fix this bug, we have two options:
- use the batch construction and the privacy budget evaluation from Abadi et al [2]. With this privacy budget evaluation based on moment accountants, we will obtain a higher privacy budget than the current privacy budget implemented in Opacus. This first option implies to re-implement the overall privacy budget evaluation.
- use the batch construction and the privacy budget evaluation from Mironov et al [1]. For this option, we have just to change the parameters of the PrivacyEngine (the batch size does not exist with this batch construction) and we have to update all examples and tutorials in order to use batches sampled according to the Sampled Gaussian Mechanism.

[1] Rényi Differential Privacy of the Sampled Gaussian Mechanism
[2] Deep Learning with Differential Privacy
[3] Improving deep learning with differential privacy using gradient encoding and denoising

examples/dcgan.py
ConstanceBeguier: As explained previously, if we would like to evaluate the privacy budget with Mironov et al evaluation [1], we must use batches sampled according to the Sampled Gaussian Mechanism. As far as I know, this batch construction cannot be directly implemented with RandomSampler or other PyTorch samplers. Thus, I suggest to add this UniformWithReplacementSampler sampler in Opacus in order to make easier the use of the correct batch construction by the Opacus users.

[1] Rényi Differential Privacy of the Sampled Gaussian Mechanism

opacus/privacy_engine.py
Darktex: ... Of course, the batch size is indeed variable. Sorry, my bad!

examples/dcgan.py
Darktex: Oh no, this comment was entirely about how to organize the API.

Whatever the sampler implementation used, currently the PrivacyEngine does not own your dataloader, which means that we leave configuring it properly entirely to the client. This was already a sore point before, because you would normally want to use a secure generator for it which is something that is easy to forget.

opacus/utils/uniform_sampler.py
Darktex, line 5: This component is almost good to go! Let's add a docstring to this so that it's well documented.

opacus/utils/uniform_sampler.py
```diff
@@ -0,0 +1,34 @@
+import torch
```
Darktex, line 1: Our repo bot will shout at us if we don't start every file with the following:

```
#!/usr/bin/env python3
# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```

opacus/privacy_engine.py
Darktex: Indeed, this question now doesn't make sense in light of the variable batch size. 

I agree to add this capability, but I want to make sure we don't introduce a breaking change all of a sudden as this is an API change.

How about we do the following? 

1. We leave batch_size and sample_size as they are. In an ideal world we would make them optional, but this would mean moving them after `noise_multiplier` and introduce yet another breaking change.
2. We add an optional parameter `dataloader = None`.
3. If the dataloader is not `None`, we print a warning that it will override the `batch_size` and the `sample_size` settings you provided.
4. Check if the dataloader has a sample_rate. If not, raise another warning that`opacus.utils. UniformWithReplacementSampler` should be used, and refer to the examples to see how to use it.

We are working on a major API refactor for Opacus 1.0, so in the future we will make this better, but this should suffice for now.

Darktex (changes requested):
Let's make these changes non-breaking and we are good to go!

ffuuugor (commented):
Thanks a lot for the contribution - this gap between privacy accounting and the actual batch sampling was annoying.
I like the proposed API change. However, I agree with Davide that we should softer on deprecating old api and switching to the new one. Even though we're still in beta, we don't want to make sudden breaking changes.

On another note, it'd be really interesting to see if the new sampler have any effect on resource consumption:
a) does it affect running time?
b) can it potentially cause memory issues? DP-SGD in really sensitive to `batch_size` and using independent sampling can lead to variance in batch_sizes, occasionally creating batches significantly larger than `sample_rate*dataset_size`. 

We probably want to have some control on sampling strategy in our 1.0 API, and answers to those questions are important to understand how it'd affect training.

opacus/privacy_engine.py
ConstanceBeguier: I have chosen to put batch_size, sample_size and sample_rate arguments optional

ConstanceBeguier:
I pushed a new commit for the retrocompatibility of former arguments in addition of warnings.

Darktex (changes requested):
@ConstanceBeguier we discussed a bit more between ourselves, and decided to go ahead with the mild breaking change of making `batch_size` optional (breaks all clients that were passing it in positionally). Sorry it took us a while to converge :D 

We are landing #136 to address this issue, could you please rebase your PR on top of Master after this lands? :)

ConstanceBeguier:
I just rebased my PR on top of master.

Darktex (approved):
Ready to go!

PR #162, opened 2021-04-05
Conversion from RDP to (eps, delta)-DP by a new Formula

mmalekzadeh:
Summary: We propose to use the state-of-the-art formula for computing `eps` in  `opacus/privacy_analysis.py` method `get_privacy_spent()`.

related to issue: https://github.com/pytorch/opacus/issues/157

Details:

Currently, the computed epsilon from RDP to (epsilon, delta)-DP is the Line 298 in `opacus/privacy_analysis.py` that is ```eps = rdp_vec - math.log(delta) / (orders_vec - 1)```.

The current line is based on `Proposition 3` in the main RDP paper (Mironov2017):
Ilya Mironov. "Renyi differential privacy." IEEE 30th Computer Security Foundations Symposium (CSF). 2017.

However, in the following paper (Balle2020) authors provide a better conversion from RDP to DP:
Borja Balle et al. "Hypothesis testing interpretations and Renyi differential privacy." International Conference on Artificial Intelligence and Statistics. PMLR, 2020.  Particullary, Theorem 21 in the arXiv version https://arxiv.org/abs/1905.09982.

As this new analysis provides us a better bound (i.e. smaller epsilons), it should be more useful to the users of Opacus.

In this pull request:

1. I have replaced Line 298 with the formula proposed in Balle2020 Theorem 21 as explained above.

2.  I have added information about this paper to the description of the method get_privacy_spent().

As the only difference of this new formula, compared to the previous one, is that it gives us smaller epsilons, and the method's functionality does not change, I believe existing unit tests should be sufficient for this change.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Docs change / refactoring / dependency upgrade

## Checklist

- [x] The documentation is up-to-date with the changes I made.
- [x] I have read the **CONTRIBUTING** document and completed the CLA (see **CONTRIBUTING**).
- [x] All tests passed, and additional code has been covered with new tests.

Issue #157, opened 2021-03-25
From RDP to (epsilon, delta)-DP


mmalekzadeh:
Hi,

There is a better conversion from RDP to (epsilon, delta)-DP than the current implementation which is a very important aspect of this library. So, here I explain it and suggest using the new formula. Unless there is something else that I am not aware of it. I appreciate your feedback.

**Explanation:**

Currently, the computed epsilon from RDP to (epsilon, delta)-DP is this line:

https://github.com/pytorch/opacus/blob/244265582bffbda956511871a907e5de2c523d86/opacus/privacy_analysis.py#L298

That is based on Proposition 3 in the main RDP paper: https://arxiv.org/abs/1702.07476

However, in the following paper (AISTATS 2019): https://arxiv.org/pdf/1905.09982.pdf
and in Theorem 21, the authors provide a better conversion from RDP to DP.

a (lazy) code for this better conversion will be something like this:
```
term_1 = np.array([math.log((alpha-1)/alpha) for alpha in orders_vec])
term_2 = np.array([math.log(alpha) for alpha in orders_vec])
eps = rdp_vec + term_1 - (math.log(delta)+term_2) / (orders_vec - 1)
```

As this new analysis provides us a better bound (i.e. smaller epsilons), it should be more useful to the users of Opacus.

Best

ilyamironov:
Mohammad,

Thank you for flagging this! Although the savings are very modest in most practical regimes, why not use the most efficient conversion available?

If you feel up for it, how about you create a pull request that implements the change? We'd be glad to review it and check it in. A warning though - as usual, the most laborious part is not the main code, but rather the tests (and readme files).

mmalekzadeh:
Thanks Ilya, for the confirmation.  I'm happy to do this pull request.

PR #167, opened 2021-04-16
PR 162 Conversion from RDP to (eps, delta)-DP by a new Formula

JohnlNguyen:
Summary:
We propose to use the state-of-the-art formula for computing eps in opacus/privacy_analysis.py method get_privacy_spent().

related to issue: #157

Details:

Currently, the computed epsilon from RDP to (epsilon, delta)-DP is the Line 298 in opacus/privacy_analysis.py that is eps = rdp_vec - math.log(delta) / (orders_vec - 1).

The current line is based on Proposition 3 in the main RDP paper (Mironov2017):
Ilya Mironov. "Renyi differential privacy." IEEE 30th Computer Security Foundations Symposium (CSF). 2017.

However, in the following paper (Balle2020) authors provide a better conversion from RDP to DP:
Borja Balle et al. "Hypothesis testing interpretations and Renyi differential privacy." International Conference on Artificial Intelligence and Statistics. PMLR, 2020. Particullary, Theorem 21 in the arXiv version https://arxiv.org/abs/1905.09982.

As this new analysis provides us a better bound (i.e. smaller epsilons), it should be more useful to the users of Opacus.

In this pull request:

I have replaced Line 298 with the formula proposed in Balle2020 Theorem 21 as explained above.

I have added information about this paper to the description of the method get_privacy_spent().

As the only difference of this new formula, compared to the previous one, is that it gives us smaller epsilons, and the method's functionality does not change, I believe existing unit tests should be sufficient for this change.

https://github.com/pytorch/opacus/pull/162

Merging this PR the correct way

Differential Revision: D27825676

### DIFF
--- a/opacus/privacy_analysis.py
+++ b/opacus/privacy_analysis.py
@@ -272,15 +272,17 @@ def get_privacy_spent(
 ) -> Tuple[float, float]:
     r"""Computes epsilon given a list of Renyi Differential Privacy (RDP) values at
     multiple RDP orders and target ``delta``.
-
+    The computation of epslion, i.e. conversion from RDP to (eps, delta)-DP,
+    is based on the theorem presented in the following work:
+    Borja Balle et al. "Hypothesis testing interpretations and Renyi differential privacy."
+    International Conference on Artificial Intelligence and Statistics. PMLR, 2020.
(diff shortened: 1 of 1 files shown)


JohnlNguyen:
PR is originally from @mmalekzadeh but we merged it the wrong way so I have to revert and merge it again

Issue #191, opened 2021-06-23
Opacus DDP integration


thomasflynn918:
Hello,

We've been working on DP training with PyTorch's Distributed DataParallel library (DDP) and we thought you might be interested. DDP enables faster training for Opacus over multiple GPUs, compared to the existing implementation for distributed training.

### Overview

The naive implementation for distributed training (e.g., this tutorial: [Writing Distributed Applications with PyTorch — PyTorch Tutorials 1.9.0+cu102 documentation](https://pytorch.org/tutorials/intermediate/dist_tuto.html#distributed-training), or the existing code in Opacus: [Multi-GPU example on CIFAR-10 (#170) · pytorch/opacus@f29110d · GitHub](https://github.com/pytorch/opacus/commit/f29110db3d408e6cf78546f342f7177b2cdb1c28)) works as follows:

- Split the data across nodes
- Do the forward pass on each node
- Do the backward pass on each node
- Clip and add noise to the gradients on each node (with either per-layer clipping or flat clipping)
- Synchronize the noisy gradients by averaging them across all nodes using `torch.distributed`'s all_reduce operator
- Take an optimizer step on each node using the synchronized gradients
  

Instead, our DDP integration with Opacus works as follows:

- Split the data across nodes
- Do the forward pass on each node
- Do the backward pass and the aggregation *in parallel*, thanks to `torch.nn.parallel.DistributedDataParallel`. During this process, the gradients *for each layer* are clipped, noised, and sent to the other nodes for aggregation *as soon as* they become available from backpropagation.  This is implemented using PyTorch's backward hook mechanism.
- Take an optimizer step on each node (on the same gradients)

Logically, our DDP integration with Opacus should be equivalent with distributed Opacus with per-layer clipping.  From a performance perspective, our implementation is much faster than distributed Opacus because it overlaps gradient computations with synchronization.
  
### Basic evaluation
Here is a quick comparison of the performance for our Opacus-DDP implementation vs Opacus' own distributed implementation.  We train a ResNet-50 over Imagenet using 32 GPUs, with a local batch size of 32 images per GPU (for a total batch size of 1024.)  The error bars represent 1 standard deviation:

![](https://i.imgur.com/X4ovgZS.png)

[...]

PR #196, opened 2021-07-06
DDP support for faster distributed training

tholop:
This PR starts to add support for PyTorch DDP as proposed in https://github.com/pytorch/opacus/issues/191 (cc @thomasflynn918 @aik7).
We ported our code to a more recent version of Opacus, but there are still some changes to discuss, in particular how to plug the DDP hook into GradSample module. 

I also fixed some bugs related to distributed training (see below) in the same PR. I'm happy to open separate PRs to merge these smaller changes separately if you prefer.

## Types of changes

More context in this issue: https://github.com/pytorch/opacus/issues/191.

This PR:
- Added a prototype for the DDP hook in `grad_sample_module.py`
- Added distributed Poisson sampling
- Updated the CIFAR10 end-to-end example (with Slurm cluster support)
- Fixed bugs in the tests and examples where all the GPUs were working
on the same data
- Fixed bugs in the naive distributed training implementation (missing
initial synchronization, attempt to do allreduce on parameters without
gradients)
- Fixed bug where DPDDP was not detected, because the privacy engine was checking `self.privacy_engine.module` instead of looking inside the GradSample module `self.privacy_engine.module._module`.

- Added tests to compare the weights with/without DDP hook

## Checklist

*(this is still work in progress, I'll update the documentation once we agree on the API)*

## TODO

- [x] Validate how the hook is plugged into the GradSample module
- [x] Add support for empty Poisson batches in the hook
- [x] Ensure that the privacy accounting is still accurate with DDP
- [x] Generate the noise only on the GPU that needs it, to save some time (especially with CSPRNG)
- [x] Support per-layer clipping with different thresholds for each layer

### DIFF
--- a/opacus/grad_sample/grad_sample_module.py
+++ b/opacus/grad_sample/grad_sample_module.py
@@ -108,7 +108,7 @@ class GradSampleModule(nn.Module):
         Special function to enable DDP support on top of a GradSampleModule
         """
 
-        self.ddp_hook_activated = True
+        self.ddp_hooks = True
 
         # We store the number of layers for the per-layer clipping
         self.n_params = 0
@@ -125,7 +125,7 @@ class GradSampleModule(nn.Module):
         """
         self.disable_hooks()
 
-        if self.ddp_hook_activated:
+        if self.ddp_hooks:
             pass
             # TODO: remove DDP hooks
 
@@ -316,7 +316,6 @@ class GradSampleModule(nn.Module):
         del p.grad_sample
 
         # Average (or sum) across the batch
-        # TODO: add support for Poisson batch sampling
         res = engine.clipper._scale_summed_grad(p.summed_grad, batch_size)
 
         del p.summed_grad
--- a/opacus/grad_sample/grad_sample_module.py
+++ b/opacus/grad_sample/grad_sample_module.py
@@ -286,7 +286,7 @@ class GradSampleModule(nn.Module):
 
     def ddp_backward_callback(self, engine, p, grad):
         """
-        This hook operates like PrivacyEngine.step(), but on a single layer:
+        This hook operates like ``PrivacyEngine.step()``, but on a single layer:
         1. clip_and_accumulate
         2. get the clip_values with clipper.pre_step()
         3. add the noise
--- a/opacus/grad_sample/grad_sample_module.py
+++ b/opacus/grad_sample/grad_sample_module.py
@@ -101,33 +101,17 @@ class GradSampleModule(nn.Module):
                 )
         self.enable_hooks()
 
-    def add_ddp_hook(
-        self, engine, loss_reduction: str = "mean", batch_first: bool = True
-    ) -> None:
-        """
-        Special function to enable DDP support on top of a GradSampleModule
-        """
-
-        self.ddp_hooks = True
-
-        # We store the number of layers for the per-layer clipping
-        self.n_params = 0
-
-        # Each layer has its own callback to clip/noise
-        for p in self.parameters():
-            if p.requires_grad:
-                self.n_params += 1
-                p.register_hook(partial(self.ddp_backward_callback, engine, p))
-
     def remove_hooks(self) -> None:
         """
         Removes hooks added by add_hooks()
         """
         self.disable_hooks()
 
-        if self.ddp_hooks:
-            pass
-            # TODO: remove DDP hooks
+        if hasattr(self, "ddp_hooks"):
+            while self.ddp_hooks:
+                handle = self.ddp_hooks.pop()
+                handle.remove()
+            delattr(self, "ddp_hooks")
 
         if not hasattr(self, "autograd_grad_sample_hooks"):
             raise ValueError("Asked to remove hooks, but no hooks found")
@@ -283,92 +267,6 @@ class GradSampleModule(nn.Module):
     def is_supported(cls, module: nn.Module) -> bool:
         """Check if this module is supported"""
         return type(module) in cls.GRAD_SAMPLERS or type(module) is DPLSTM
-
-    def ddp_backward_callback(self, engine, p, grad):
-        """
-        This hook operates like ``PrivacyEngine.step()``, but on a single layer:
-        1. clip_and_accumulate
-        2. get the clip_values with clipper.pre_step()
-        3. add the noise
-        """
-
-        # Get the norm of the gradient for each sample for this layer
-        all_norms = calc_sample_norms(
-            named_params=((None, p.grad_sample),),
-            flat=False,
-        )
-
-        # Get the constant clipping factor for a single layer
-        clipping_factor = engine.clipper.norm_clipper.calc_clipping_factors(all_norms)
-        assert len(clipping_factor) == 1
-        clip_factor = clipping_factor[0]
-        batch_size = p.grad_sample.shape[0]
-
-        # Do the clipping
-        summed_grad = engine.clipper._weighted_sum(clip_factor, p.grad_sample)
-
-        # accumulate the summed gradient for this mini-batch
-        if hasattr(p, "summed_grad"):
-            p.summed_grad += summed_grad
-        else:
-            p.summed_grad = summed_grad
-
-        del p.grad_sample
-
-        # Average (or sum) across the batch
-        res = engine.clipper._scale_summed_grad(p.summed_grad, batch_size)
-
-        del p.summed_grad
-
-        # NOTE: `self.clipper.pre_step()` returns a tensor with n=n_params clip_values
-        # where each clip_value is the L2 norm of `threshs`:
-        #   `max_norm = threshs.new_full((n,), threshs.norm(2))`
-        # With per layer clipping, `threshs` is a tensor with n flat_values
-        assert len(engine.clipper.norm_clipper.thresholds) == 1
-        clipping_thresh = engine.clipper.norm_clipper.thresholds[0]
-        clip_value = (self.n_params ** 0.5) * clipping_thresh
-        # if engine.rank == 0:
-        #     print(
-        #         f"max_grad_norm: {engine.max_grad_norm} \nclipping_tresh: {clipping_thresh}\nclip_value: {clip_value}"
-        #     )
-
-        # Only one GPU adds noise
-        if engine.rank == 0:
-            noise = _generate_noise_ddp(engine, clip_value, res)
-            if engine.loss_reduction == "mean":
-                noise /= batch_size
-            res += noise
-
-        return res
-
-
-def _generate_noise_ddp(
-    engine, max_grad_norm: float, reference: nn.parameter.Parameter
-) -> torch.Tensor:
-    r"""
-    Generates a tensor of Gaussian noise of the same shape as ``reference``.
-
-    The generated tensor has zero mean and standard deviation
-    sigma = ``noise_multiplier x max_grad_norm ``
-
-    Args:
-        max_grad_norm : The maximum norm of the per-sample gradients.
-        reference : The reference, based on which the dimention of the
-            noise tensor will be determined
-
-    Returns:
-        the generated noise with noise zero and standard
-        deviation of ``noise_multiplier x max_grad_norm ``
-    """
-    if engine.noise_multiplier > 0 and max_grad_norm > 0:
-        return torch.normal(
-            0,
-            engine.noise_multiplier * max_grad_norm,
-            reference.shape,
-            device=engine.device,
-            generator=engine.random_number_generator,
-        )
-    return torch.zeros(reference.shape, device=engine.device)
 
 
 def _get_batch_size(
--- a/opacus/grad_sample/grad_sample_module.py
+++ b/opacus/grad_sample/grad_sample_module.py
@@ -333,12 +333,11 @@ class GradSampleModule(nn.Module):
         #         f"max_grad_norm: {engine.max_grad_norm} \nclipping_tresh: {clipping_thresh}\nclip_value: {clip_value}"
         #     )
 
-        noise = _generate_noise_ddp(engine, clip_value, res)
-        if engine.loss_reduction == "mean":
-            noise /= batch_size
-
         # Only one GPU adds noise
         if engine.rank == 0:
+            noise = _generate_noise_ddp(engine, clip_value, res)
+            if engine.loss_reduction == "mean":
+                noise /= batch_size
             res += noise
 
         return res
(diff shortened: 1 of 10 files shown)


opacus/grad_sample/grad_sample_module.py
```diff
@@ -260,6 +298,94 @@ def is_supported(cls, module: nn.Module) -> bool:
...
+        """
+
+        # Get the norm of the gradient for each sample for this layer
+        all_norms = calc_sample_norms(
+            named_params=((None, p.grad_sample),),
+            flat=False,
+        )
+
+        # Get the constant clipping factor for a single layer
+        clipping_factor = engine.clipper.norm_clipper.calc_clipping_factors(all_norms)
```
alexandresablayrolles, line 316: Is the max_grad_norm the same across layers?

opacus/privacy_engine.py
```diff
@@ -155,6 +159,10 @@ def __init__(
 
         self.module = GradSampleModule(module)
 
+        # TODO: add the hooks later? Check which parameters are accessed
+        if isinstance(module, torch.nn.parallel.DistributedDataParallel):
+            self.module.add_ddp_hook(engine=self)
```
alexandresablayrolles, line 164: Only do this if clip_per_layer=True I guess?

alexandresablayrolles (changes requested):
Thanks for the great diff! I left a couple of comments. From looking at the code it looks like there are still some TODOs here and there, are you expecting to update the PR with the corresponding implementations?

opacus/grad_sample/grad_sample_module.py
```diff
@@ -260,6 +298,94 @@ def is_supported(cls, module: nn.Module) -> bool:
...
+        # Only one GPU adds noise
+        if engine.rank == 0:
+            res += noise
+
+        return res
+
+
+def _generate_noise_ddp(
+    engine, max_grad_norm: float, reference: nn.parameter.Parameter
+) -> torch.Tensor:
```
ffuuugor, line 363: Point to discuss.
Is there any way to decouple this from the core GradSampleModule? The idea behind recent refactoring, which introduced GradSampleModule, was exactly to avoid the coupling between per sample gradient computation, clipping and noise generation.
This change seems to reverse this. Although I understand that doing clipping & noise within the backward hook is the very core idea of this change, so it might be tricky

opacus/grad_sample/grad_sample_module.py
```diff
@@ -100,6 +102,42 @@ def add_hooks(self, loss_reduction: str = "mean", batch_first: bool = True) -> N
...
+        #             )
+        #         )
+
+        # We iterate over the parameters and not the submodules
+        for p in self.parameters():
+            if p.requires_grad:
+                self.n_params += 1
+                p.register_hook(partial(self.ddp_backward_callback, engine, p))
+
     def remove_hooks(self) -> None:
```
ffuuugor, line 141: Do we need to update `remove_hooks` to account for new hooks for the proper cleanup?

opacus/tests/ddp_hook_test.py
```diff
@@ -0,0 +1,278 @@
...
+                noise_multiplier,
+                max_grad_norm,
+            ),
+            nprocs=world_size,
+            join=True,
+        )
+    return world_size
+
+
+class GradientComputationTest(unittest.TestCase):
```
ffuuugor, line 235: Since you've essentially re-implemented clipping and adding noise, can you add specific tests for this too? Or instead update existing `per_sample_gradient_clip_test.py` and `privacy_engine_test.py` to include distributed case.

ffuuugor (changes requested):
Thanks a lot for your PR!
Overall looks really good, I've added some comments on a few aspects of the design.

Also, just to clarify - do you plan to address TODO items later in this PR or leave them for later (both options are fine, just want to be clear on the scope)

opacus/grad_sample/grad_sample_module.py
```diff
@@ -260,6 +298,94 @@ def is_supported(cls, module: nn.Module) -> bool:
...
+
+        del p.summed_grad
+
+        # NOTE: `self.clipper.pre_step()` returns a tensor with n=n_params clip_values
+        # where each clip_value is the L2 norm of `threshs`:
+        #   `max_norm = threshs.new_full((n,), threshs.norm(2))`
+        # With per layer clipping, `threshs` is a tensor with n flat_values
+        assert len(engine.clipper.norm_clipper.thresholds) == 1
+        clipping_thresh = engine.clipper.norm_clipper.thresholds[0]
+        clip_value = (self.n_params ** 0.5) * clipping_thresh
```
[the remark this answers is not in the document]
tholop: I agree that this snippet is a bit confusing. However, it seems that this behavior (increasing the multiplier by `sqrt(n_params)`) is the current behavior when we use `_Dynamic_Clipper_` with `clip_per_layer=True` and `max_grad_norm=C` for some float C. 
What do you think @thomasflynn918? 

In any case, I will try to make this part clearer when I refactor the hook -- especially if we have a different threshold for each layer.

opacus/grad_sample/grad_sample_module.py
tholop: Yes, for now -- but I plan to add support for per-layer thresholds

opacus/privacy_engine.py
tholop: Yes, indeed! I fixed this

opacus/grad_sample/grad_sample_module.py
tholop: Yes, it would be neater to decouple the clipping/noising. I thought about it, and I can try to divide this piece into two parts:
- the per-layer clipping/noising hooks (`ddp_backward_callback` and utils such as `_generate_noise_ddp`) would be defined in PrivacyEngine, and called by the hook helpers
- the hook helpers (add/remove ddp hook) would stay in GradSampleModule

In addition to being neater, this structure might make easier to modify the DDP functionality (e.g., per-layer thresholds) without having to change GradSampleModule. @ffuuugor what do you think?

Another option would be to pack the DDP functions into a separate module, similarly to `dp_ddp.py`

opacus/grad_sample/grad_sample_module.py
tholop: Indeed, I'll add a helper once we clarify the architecture

opacus/tests/ddp_hook_test.py
tholop: Will do! I'll also try to reuse more code if possible

opacus/privacy_engine.py
```diff
@@ -475,6 +504,7 @@ def step(self, is_empty: bool = False):
             if self.loss_reduction == "mean":
                 noise /= batch_size
 
+            # TODO: Why do we need to generate noise on every GPU if only rank0 will use it?
```
tholop, line 507: @alexandresablayrolles is there any reason for generating the noise on every GPU, even if only one GPU is using it? 
It seems a bit like a waste of time (especially for large networks with the CSPRNG turned on).

tholop:
Hi @ffuuugor and @alexandresablayrolles, thanks for the feedback! I addressed your comments inline and pushed a few minor changes. 

Regarding the TODOs: I would like to address them in this PR. I just wanted to get some feedback on the overall structure first before starting to work on the rest.

I am going to refactor the hook and try to reuse more clipping functions. I'll ping you with a re-request when I'm done with more significant changes!

opacus/grad_sample/grad_sample_module.py
```diff
@@ -260,6 +284,94 @@ def is_supported(cls, module: nn.Module) -> bool:
         """Check if this module is supported"""
         return type(module) in cls.GRAD_SAMPLERS or type(module) is DPLSTM
 
+    def ddp_backward_callback(self, engine, p, grad):
+        """
+        This hook operates like PrivacyEngine.step(), but on a single layer:
+        1. clip_and_accumulate
+        2. get the clip_values with clipper.pre_step()
+        3. add the noise
```
ashkan-software, line 292: likewise, here :)

opacus/grad_sample/grad_sample_module.py
```diff
@@ -260,6 +284,94 @@ def is_supported(cls, module: nn.Module) -> bool:
         """Check if this module is supported"""
         return type(module) in cls.GRAD_SAMPLERS or type(module) is DPLSTM
 
+    def ddp_backward_callback(self, engine, p, grad):
+        """
+        This hook operates like PrivacyEngine.step(), but on a single layer:
```
ashkan-software, line 289:
```suggestion
        This hook operates like ``PrivacyEngine.step()``, but on a single layer:
```

opacus/grad_sample/grad_sample_module.py
```diff
@@ -260,6 +284,94 @@ def is_supported(cls, module: nn.Module) -> bool:
...
+    engine, max_grad_norm: float, reference: nn.parameter.Parameter
+) -> torch.Tensor:
+    r"""
+    Generates a tensor of Gaussian noise of the same shape as ``reference``.
+
+    The generated tensor has zero mean and standard deviation
+    sigma = ``noise_multiplier x max_grad_norm ``
+
+    Args:
+        max_grad_norm : The maximum norm of the per-sample gradients.
```
ashkan-software, line 357: engine is also an input, so please add here

opacus/grad_sample/grad_sample_module.py
```diff
@@ -100,11 +101,34 @@ def add_hooks(self, loss_reduction: str = "mean", batch_first: bool = True) -> N
                 )
         self.enable_hooks()
 
+    def add_ddp_hook(
+        self, engine, loss_reduction: str = "mean", batch_first: bool = True
+    ) -> None:
+        """
+        Special function to enable DDP support on top of a GradSampleModule
```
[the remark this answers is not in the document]
tholop: Will do, thanks for the feedback!

opacus/privacy_engine.py
tholop: Fixed.

opacus/grad_sample/grad_sample_module.py
ffuuugor: Oops, sorry for the delay
I like the idea of moving DP-specific functionality (e.g. `_generate_noise_ddp`) to PrivacyEngine. Do I understand correctly, that you'll implement the hook that does noise and clipping somewhere else, and will pass it to `add_ddp_hook` in GradSampleModule?

The principle we want to adhere to is GradSampleModule should make sense outside of DP context as well

opacus/grad_sample/grad_sample_module.py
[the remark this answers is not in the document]
ffuuugor: I think you're indeed right, that `_Dynamic_Clipper_` does the same. But to me that's more of an argument to change `_Dynamic_Clipper_` too - for the reasons I've outlined above I think it's more consistent if by `max_grad_norm` we always mean "L2 norm of the entire grad vector" (unless it's a list, then it's reasonable to expect different interpretation)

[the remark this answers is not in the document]
tholop: Good point. Since I want to support per-layer clipping in the general case, I am now expecting `max_grad_norm` to be a list. I dropped support for `_Dynamic_Clipper_` because it was only making sense in the cases where it was equal to `ConstantPerLayerClipper` (we can still implement distributed versions of the dynamic schemes later).

opacus/grad_sample/grad_sample_module.py
tholop: Exactly -- with a small change: when I implemented this design, I found it more convenient to define and register the hooks directly from the PrivacyEngine. I still give a list of handles to GradSampleModule, to make it easier to remove the hooks later.

I think that I can go even further and completely leave GradSampleModule untouched, if I implement `remove_ddp_hooks` into the PrivacyEngine directly. @ffuuugor please let me know if you would like me to do that, after you've had a look at the new code

tholop:
Hi, I refactored the DDP hook and made some modifications based on your feedback. Here is the TL;DR of my commits:
- The DDP hook now lives with the PrivacyEngine
- Support for constant per-layer clipping with different thresholds for each layer
- Support for Poisson batches and proper accounting 

 I don't plan to add new features in this PR, so I'm ready to clean-up and address any comments you might have. Thanks!

opacus/utils/tensor_utils.py
```diff
@@ -42,6 +42,34 @@ def calc_sample_norms(
...
+
+    Example:
+        >>> t1 = torch.rand((2, 5))
+        >>> calc_sample_norms_one_layer(t1)
+            tensor([1.4757, 0.8128])
+
+    Returns:
+        A tensor of norms
+    """
+    norms = param.view(len(param), -1).norm(2, dim=-1)
```
alexandresablayrolles, line 69: Nit: replace len() with .size(0) since it is a Tensor?

tholop: Why not -- I was following the same structure as `calc_sample_norms`, which is using `len()`on a tensor:
```python
norms = [param.view(len(param), -1).norm(2, dim=-1) for name, param in named_params]
```

 @alexandresablayrolles should I modify the original `calc_sample_norms` to use `size(0)` too then?

tholop:
Hi @alexandresablayrolles et al, just checking -- is there anything else I can do to help merging this PR?

alexandresablayrolles:
@tholop sorry for the delay, I will take a look beginning of next week.

alexandresablayrolles:
Hi @tholop,
I think the diff is in great shape. I have a couple of additional requests:
- It seems that the cifar10 example currently cannot accommodate per layer thresholds, can you change that? This is particularly important as it will also give us a better idea of the privacy/accuracy trade-off we get with the per-layer DDP
- Can you refactor the _generate_noise so that it always takes a grad? I think your new function is better and should replace the older one instead of both existing at the same time.

If you can do this, I will then merge the diff into Opacus. Thanks again for the great diff!

tholop:
Hi @alexandresablayrolles, thanks for the review! 

Indeed, I forgot to update the cifar10 example with my last changes. I can add an example for a simple per-layer clipping where each layer has `max_layer_grad_norm = args.max_per_sample_grad_norm / sqrt(n_layers)`. I think that other policies are possible (e.g., assign different weights to different layers), but I don't have a clear understanding of the privacy/accuracy tradeoff yet and it is probably out of scope for the cifar10 example.

I'll ping you with the changes by the end of this month.

PR #212, opened 2021-09-29
[new API] unit tests back to green

ffuuugor:
Updating unit tests with the new API.
This mostly focuses on getting the existing tests green. More work is planned for the comprehensive tests for the new components.

## Disabled tests
Some tests make sense, but refer to the component not yet implemented in the new version. Those are temporary disabled - the plan is to bring them back as we implement more stuff. This includes:

- Docstring examples for `PrivacyEngine`. Pending `PrivacyEngine` docstirngs
- secure rng tests
- model validation tests
- virtual steps 

## Modified tests
Some tests we didn't make much sense, but then high-level idea behind them is good. Those were replaced.

### Poisson test. 
In the previous implementation it was checking that training doesn't fail with `UniformWithReplacementSampler`. Now this is on by default and therefore test now belongs to `PrivacyEngine` - in fact the same check is now covered by `PrivacyEngine_test.test_basic`. 
However, we didn't have a unit test for the `UniformWithReplacementSampler` itself, so `poisson_test.py` now checks basic behaviour of the uniform sampler

### PrivacyEngine test
Over a few cycles of development this test suite evolved into something very complex. Initially it was running two models (private and vanilla) in the `setUp` method and comparing the results in the test cases. Over time this started to fall short of our needs and now many test cases run models on top of that, ignoring the `setUp` runs. 
This breaks the conceptual simplicity - it was really hard to understand what's going on.

In this PR we:
- simplify the model running: models initiated and trained in every test case as needed, no `setUp`
- All private/vanilla checks are now encapsulated in `test_compare_to_vanilla`; We check 4 combinations of noise/no_noise, clip/no_clip and make sure gradients and weights match if there's no noise or clipping and don't match otherwise.
- Few test cases are removed:
    - `test_throws_on_bad_per_layer_maxnorm_size`, `test_grad_matches_original_per_layer_clipping` - only make sense for per-layer clipping, will probably bring back when we get to implementing it
    - `test_throws_double_attach` / `test_attach_detach_attach` - multiples calls to `PrivacyEngine` are allowed now. Rest should move to `GradSampleModule`
    - `test_privacy_analysis_epsilon` - will be covered in accountant specific tests
    - `test_deterministic_noise_generation` - covered by `test_deterministic_run`
    - `test_sampling_rate_less_than_one` - this is now computed automatically

## Integration tests
I've also updated all example scripts with the new API. The only exception is cifar10 example, which relies heavily on DDP and per-layer clipping - it wouldn't make much sense to touch it before we support DDP.

Couple notes:
- sample_rate is replaced with batch_size as a script parameter. I believe batch_size is more intuitive and easier for most users to work with. Fun fact: when we introduced sample_rate and replaced batch_size, the new equivalent batch_sizes were all over the place (e.g. in circle ci config, mnist integration test, batch_size 32 was replaced with sample rate 0.004, equivalent of batch_size=200)
- secure generators are removed from the client code, but will be introduced back as part of `secure_mode` implementation


karthikprasad:
Thanks for this. The changes LGTM. Do you plan to also fix the failing integration test in this PR?

alexandresablayrolles (approved):

PR #213, opened 2021-09-30
[new API] Scheduler

alexandresablayrolles:
[WIP] Initial implementation for the noise scheduler. 

The idea is to have an abstract class _NoiseScheduler, that all schedulers will extend. This is similar to what Pytorch does for learning rate schedules. 

Setting the noise level itself is very simple (directly modify `optimizer.noise_multiplier`), the accountant should gracefully handle the rest (i.e. the noise will get added to the accountant.steps during the optimizer.step).

What is left to do in this diff:
- Add unit tests
- Implement basic schedules (steps, exponential)
- Add checks (e.g. check that there is at least one optimizer step between two changes of the noise multiplier, or else output a warning)


opacus/scheduler.py
```diff
@@ -0,0 +1,83 @@
...
+        raise NotImplementedError
+
+    def step(self):
+        self.last_epoch += 1
+        noise_multiplier = self.get_noise_multiplier()
+        self.optimizer.noise_multiplier = noise_multiplier
+
+
+class ExponentialNoise(_NoiseScheduler):
+    def __init__(self, optimizer: DPOptimizer, gamma: int, last_epoch: int = -1):
```
ffuuugor, line 44: should `gamma` be int?

opacus/scheduler.py
```diff
@@ -0,0 +1,83 @@
+from typing import Callable
+
+from .optimizer import DPOptimizer
+
+
+class _NoiseScheduler(object):
+    def __init__(self, optimizer: DPOptimizer, last_epoch=-1):
+        # Attach optimizer
+        if not isinstance(optimizer, DPOptimizer):
+            raise TypeError("{} is not a DPOptimizer".format(type(optimizer).__name__))
```
ffuuugor, line 10: Do we want to do that on top of the type hint?

opacus/scheduler.py
```diff
@@ -0,0 +1,83 @@
...
+    def __init__(
+        self, optimizer: DPOptimizer, step_size: int, gamma: float, last_epoch: int = -1
+    ):
+        self.step_size = step_size
+        self.gamma = gamma
+        super().__init__(optimizer, last_epoch)
+
+    def get_noise_multiplier(self):
+        # Only change noise_multiplier when at a 'step'
+        if self.last_epoch == 0 or self.last_epoch % self.step_size == 0:
```
ffuuugor, line 80:
```suggestion
        if self.last_epoch == 0 or self.last_epoch % self.step_size != 0:
```

opacus/tests/scheduler_test.py
```diff
@@ -0,0 +1,52 @@
...
+        step_size = 2
+        scheduler = StepNoise(self.optimizer, step_size=step_size, gamma=gamma)
+
+        self.assertTrue(self.optimizer.noise_multiplier == 1.0)
+        scheduler.step()
+        scheduler.step()
+        self.assertTrue(self.optimizer.noise_multiplier == gamma)
+        scheduler.step()
+        scheduler.step()
+        self.assertTrue(self.optimizer.noise_multiplier == gamma ** 2)
```
ffuuugor, line 44: I think we should add extra asserts in between steps: it's pretty fascinating coincidence, that this exact setup (step_size=2, only check after two steps) didn't catch the bug from the comment above

ffuuugor (approved):
Looks great: clean, simple, takes all the best parts from `LRScheduler`

Approving now, but please take a look at the line comments before landing

opacus/scheduler.py
alexandresablayrolles: Hmm but the hint is not enforced right?

ffuuugor: I don't really have strong opinion on this - I see an appeal of having a user-friendly message for a potentially common mistake too.

I personally wouldn't have added it: after all, we don't check and enforce other parameters. And if someone isn't using IDE or mypy to highlight the typing issues, they've probably brought this upon themselves.

Note, that if it's not for this message, the code will still fail with "optimizer has no attiribute noise_multiplier" couple lines later - so it's only a matter of a different message

PR #216, opened 2021-10-06
Pass down batch_first parameter

ffuuugor:
As noticed by @samdow, grad_samples doesn't look right when using `PrivacyEngine` with `batch_first=False`.

The problem is, since `GradSampleModule` was introduced, we don't pass down `batch_first` from `PrivacyEngine` to `GradSampleModule` (see `privacy_engine:162-164`)

In addition to that, none of our tests caught that: grad_sample tests use `GradSampleModule` directly and bypass PrivacyEngine, privacy_engine tests don't check `batch_first` case.

This PR fixed both issues:
- We initialize `GradSampleModule` with full set of parameters 
- in `privacy_engine_test.py` we introduce additional set of test cases: we now run all tests for two models: ConvNet (as before) and NLP model (Embedding + MultiheadAttention).
- We also add one more check to `privacy_engine_test.py`: `test_grad_sample_shape`. Surprisingly, even with the new NLP model, none of the existing test cases caught `batch_first` issue; when you aggregate over the first dimension, you get correct result despite having set `batch_first` wrong

### DIFF
--- a/opacus/layers/dp_multihead_attention.py
+++ b/opacus/layers/dp_multihead_attention.py
@@ -41,7 +41,7 @@ class SequenceBias(nn.Module):
 
     def _reset_parameters(self):
         r"""
-        assing's Normally distributed random values to bias.
+        assigns Normally distributed random values to bias.
         """
         nn.init.normal_(self.bias)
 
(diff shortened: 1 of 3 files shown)


karthikprasad:
Ouch! Great catch and fix!

PR #219, opened 2021-10-08
[New API] Virtual step

igorshh:
## Background

We want to support virtual steps, while requiring minimal (ideally none at all) input from a user. 
We want to get as close as possible to vanilla PyTorch API - in vanilla PyTorch you can simply call `forward+backward` multiple times before `optimizer.step()` and gradients would simply accumulate into `p.grad` (effectively simulating larger batch size).

With DP-SGD we have a few problems to solve before we can do the same:
1) Accountant needs to know how many batches you've processed.
2) Optimizer also needs to know how many batches you've processed to average gradients with the correct denominator
3) Simply accumulating grad_samples between steps doesn't make much practical sense - you don't win much extra memory if you still have to store full per-sample gradients for the large batch. And we can't really aggregate grad_samples before we've clipped them
4) We also need to force `zero_grad` between virtual steps: `grad_sample` can't be used twice, it breaks DP guarantees.

## New behaviour

- We change the way grad_samples are computed. Before we've manually set some layers as recurrent and would use "create_or_accumulate" method there, while all standard modules used "create_or_expand". This creates at least one problem: if you're using `GradSampleModule` separately and call forward/backward multiple times in a row, you'd get incorrect grad samples for LSTM - different batches would be accumulated as if it's the same batch called recurrently. Instead now we:
    - Accumulate by default in a temporary tensor
    - In a backward hook check if it's the last backward in the stack (by checking that activations are empty)
    - If it's the last backward for that layer, we put temporary tensor into `p.grad_sample`, possibly expanding it
    - This way we accumulate within a batch and stack across batches, with a common approach across all layers
- In between batches grad samples are accumulated into a list. This is better than the current approach of continuously stacking them into a single tensor, because also exposes information about individual batches: you know how many batches of which size has produced these per sample gradients
- introducing `BatchMemoryManager` - way to decouple physical and logical batch size (again in 1 line of code). Example usage:

```
model, optimizer, data_loader = ...
with BatchMemoryManager(data_loader=data_loader, optimizer=optimizer, max_physical_batch_size=3) as data_loader:
    for x, y in data_loader:
        optimizer.zero_grad()
        out = model(x)
        loss = criterion(out, y)
        loss.backward()
        optimizer.step()
```

The code will split the larger batched produced by the original `data_loader` into smaller chunks, while also telling `optimizer` when to skip `step()`. A slightly counter-intuitive bit here is that `optimizer.step()` sometimes will be a no-op.

The good thing about `BatchMemoryManager` is that it works both for preventing OOMs caused by Poisson sampling and to save memory using virtual steps.

NB: calling forward/backward multiple times before calling `optimizer.step()` without `BatchMemoryManager` is allowed, but is memory inefficient.

## Accumulation summary
- When using `GradSampleModule` separately: ALLOWED ✅ 
- When using `PrivacyEngine` with Poisson sampling off: ALLOWED ✅ 
- When using `PrivacyEngine` with Poisson sampling off: PROHIBITED 🛑 

## Style stuff
I've made some cosmetic changes to some interfaces - it's all down to individual taste, and I'm very open to rolling it back if someone disagrees:
- `grad_sampler_fn` now returns grad_sample tensors instead of updating p.grad_sample. 
- Removed unused parameter `batch_dim` from grad_samplers

### DIFF
--- a/opacus/grad_sample/grad_sample_module.py
+++ b/opacus/grad_sample/grad_sample_module.py
@@ -326,18 +326,18 @@ class GradSampleModule(nn.Module):
         # TODO: that's not right; DPLSTM isn't always batch_first=False
         batch_dim = 0 if batch_first or type(module) is LSTMLinear else 1
 
-        A = module.activations.pop()
+        activations = module.activations.pop()
 
         if not hasattr(module, "max_batch_len"):
             # For packed sequences, max_batch_len is set in the forward of the model (e.g. the LSTM)
             # Otherwise we infer it here
-            module.max_batch_len = _get_batch_size(module, A, batch_dim)
+            module.max_batch_len = _get_batch_size(module, activations, batch_dim)
 
         n = module.max_batch_len
         if loss_reduction == "mean":
-            B = backprops * n
+            backprops = backprops * n
         elif loss_reduction == "sum":
-            B = backprops
+            backprops = backprops
         else:
             raise ValueError(
                 f"loss_reduction = {loss_reduction}. Only 'sum' and 'mean' losses are supported"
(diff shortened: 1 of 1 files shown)


ffuuugor:
cc @romovpa (re-request review button doesn't work for you for some reason)

Issue #206, opened 2021-09-20
GRU support for Opacus


ashkan-software:
## 🚀 Feature
GRU support for Opacus

## Motivation

Opacus currently does not support [Gated Recurrent Unit (GRU)](https://en.wikipedia.org/wiki/Gated_recurrent_unit). Looks like some of the members of the community are intersted in such feature. See, for example, [issue 205](https://github.com/pytorch/opacus/issues/205)

## Additional context

Supporting GRU in Opacus is a similar effort like [supporting LSTM](https://github.com/pytorch/opacus/commit/2ddd270d3ff637be6cd59c9f1e6cd860b29a6c5a) in Opacus. As maintainers of Opacus we didn't implement GRU so far mostly because nobody asked and there was no need. However, this is a good feature. We hope that one day this feature jumps to the top of our priorities so we can implement it. Or even better, maybe a contributor can send us a PR for this and we gladly work with them to support it! :)

romovpa:
Hi, I'm happy to help with this.

ffuuugor:
Good luck!

ashkan-software:
Thanks @romovpa  for doing this! 

CC @anik123

anik123:
Thanks @romovpa it would be a great help!

romovpa:
There is some progress [here](https://github.com/romovpa/opacus/pull/1). 
I've created dirty but working replacements for `nn.GRU` and `nn.RNN`. Some work is to be done to merge LSTM code with RNN/GRU (presence of the cell state `c_n` in LSTM makes it tricky), update all tests. Then I'll update examples/tutorials in a separate diff.

We have `DPGRU` and `DPRNN` layers, the parametrized part of which is based on `RNNLinear` (former `LSTMLinear`).  
For `RNNLinear` we use aggregating grad sampler which sums out the sequence dimension. My understanding is that this is enough for PrivacyEngine to work correctly. Correct me here if wrong, please.

ffuuugor:
Wow, that is some seriously impressive work! 
I've skimmed through your draft and on a high level it looks right about how it should be looking.

It's great that you've managed to bring together some of the common parts with DPLSTM implementation under the same base class (at least for the cells)

> My understanding is that this is enough for PrivacyEngine to work correctly.

That is my understanding as well: using only `RNNLinear` should do the trick. Anyways, the best way to be certain about this is to expand grad sampler test (`opacus/tests/grad_samples/dp_lstm_test.py`)

[...]

PR #222, opened 2021-10-12
Support of GRU and RNN. Refactored LSTM implementation.

romovpa:
## Types of changes

- [x] Docs change / refactoring / dependency upgrade

The original issue is to implement GRU support for opacus #206. 
This requires creating a drop-in replacement for `nn.GRU` similar to `DPLSTM` (a replacement for `nn.LSTM`). Both LSTM and GRU (as well as simple RNN) use `nn.Linear` as the only parametrized building block, so we will reuse the special version of it equipped with accumulating grad sampler (called `LSTMLinear` in the code).

## Changes

The idea of this PR is to create a set of classes with exactly the same interface as in [`torch.nn.modules.rnn`](https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/rnn.py).

1. LSTMLinear (the basic building block of DPLSTM) renamed to more generic `RNNLinear`
2. Added `DPRNNCellBase` class and DP**RNN**Cell, DP**GRU**Cell, DP**LSTM**Cell derived from it. 
3. Added `DPRNNBase` class and DP**RNN**, DP**GRU**, DP**LSTM** derived from it. Supported by grad sampler.
4. In `char-lstm-classification.py` example added flag `--mode {rnn,gru,lstm}` to choose between different models
5. `ParamRenamedModule` is replaced with `RenameParamsMixin` trait
6. `compute_seq_lengths` is moved to `opacus.utils.packed_sequences` as it was double implemented, it is useful and has clear interface.

All classes implemented in `dp_rnn` have identical interface and `state_dict` with those from `torch.nn.rnn`.

The only thing that is not implemented: the case with `proj_size > 0` (it wasn't implemented before). It can be a new issue for external contributors which is much smaller and easier for the newcomers. 

## How Has This Been Tested

- Updated tests `dp_layers/dp_rnn_test.py` and `grad_samples/dp_rnn_test.py` to cover the new features.
- Pytest passes
- Modified `examples/char-lstm-classification.py` to switch between RNN/GRU/LSTM. Run all models and checked learning curves.

Lints:
- Applied `black` reformatting
- Applied `isort`
- Checked flake8

## Checklist

### DIFF
--- a/opacus/grad_sample/utils.py
+++ b/opacus/grad_sample/utils.py
@@ -69,7 +69,6 @@ def create_or_accumulate_grad_sample(
             shape as ``param`` with extra batch dimension
     """
 
-    print(param, grad_sample, layer, layer.max_batch_len)
     if hasattr(param, "grad_sample"):
         param.grad_sample[: grad_sample.shape[0]] += grad_sample
     else:


ffuuugor:
Hey @romovpa 
This change was indeed intended for `main` branch - it fits into both versions of the opacus API, and there's no reason not to land in into the `main` now.

`experimental_v1.0` lives as an active development branch, while we're actively working on the new API. We occasionally merge the upstream changes from `main` to keep up and make the final merging easier. Now we're a bit behind on the merging from main, because we're about to land a large diff with DDP support into experimental and the merge will be painful.

When we're ready (hopefully early Nov) we'll merge experimental into main and mark it as 1.0 release.

One quirky difference between merging into main vs other branches is that for main we need to use fb infra, while for all other branches we can merge github PRs as normal.
I've linked the #206 issue, but it won't automatically close it upon merging, because PRs into main are landed through fb infra: all the commits and authorship is preserved, but the PR is marked as closed instead of merged

PR #227, opened 2021-10-18
[New API] Module Validator

karthikprasad:
## Background
New API should provide robust yet flexible ways to do model validation.
The high level goal for this module is twofold:
1. Ensure our privacy analysis is applicable to the model and training process
2. Catch failures early, provide human-friendly error messages
(Some checks, if not performed, will have lead to failures down the road, but some won't affect training, while not actually providing claimed privacy guarantees)

For a simple use case it should perform the same set of check we currently have in `DPModelInspector`:
1. we have grad_sampler for all trainable parametrized sub-modules
2. model doesn't have batch_norm layers
3. other norm layers don't track running stats
4. ~conv2d layers have predefined number of groups~
5. LSTM layers are replaced with DPLSTM
6. Model is in training  mode.

## Guiding principles
1. Have a modularised validation API instead of a long central list of validation.
2. GradSampleModule should only concern itself with per-sample gradients and not worry about DP validation.
3. Throw ALL errors instead of just throwing and stopping on the first error.
4. Errors should be as descriptive as possible.
5. Follow the existing design pattern (similar to grad samplers) 
6. Derive from and incorporate the comments on #131.

## This PR
* Allows anyone to register their custom module validators and fixers (similar to grad samplers.)
* The privacy engine makes a call to `fix` (if flag is set) and `validate` as the first steps of preparing the model for private training.

## TODO
- [x] Add tests
- [ ] Resolve TODOs in the code.
- [x] Delete the current version of model_inspector and model_modifier

### DIFF
--- a/opacus/grad_sample/grad_sample_module.py
+++ b/opacus/grad_sample/grad_sample_module.py
@@ -330,7 +330,7 @@ class GradSampleModule(nn.Module):
 
     @classmethod
     def is_supported(cls, module: nn.Module) -> bool:
-        """Check if this individu module is supported"""
+        """Check if this individual module is supported"""
         return type(module) in cls.GRAD_SAMPLERS or type(module) is DPLSTM
 
     @classmethod
--- a/opacus/validators/batch_norm.py
+++ b/opacus/validators/batch_norm.py
@@ -1,12 +1,22 @@
 #!/usr/bin/env python3
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
+import logging
+import sys
 from typing import Union
 
 import torch.nn as nn
 
-from .errors import ShouldReplaceModuleError
+from .errors import ShouldReplaceModuleError, UnsupportableModuleError
 from .utils import register_module_validator, register_module_fixer
+
+logging.basicConfig(
+    format="%(asctime)s:%(levelname)s:%(message)s",
+    datefmt="%m/%d/%Y %H:%M:%S",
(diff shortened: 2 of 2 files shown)


ffuuugor:
> Providing an explicit API might send a message to all users that validation is optional. Do you think otherwise?

That's a good point - that's not the signal we want to send. 
Given that you plan to incorporate global `default_validator`, what's your vision for the use of custom validators? Do people replace the attribute after PrivacyEngine initialization?

I think it's a good plan - probably a good balance between "easy to do" and "not obvious enough". And I think it'd be a good for us to have a tutorial on advanced features of opacus like this one

> How about adding another method is_dp_compatible <...>

~~The only somewhat strong opinion I have on this is that I'd prefer users to have to explicitly call a method that changes the architecture of their model.~~
I first wrote this, but then realised that changing LSTM to DPLSTM is not a big deal, as it's merely an implementation detail and has no effect on the training. But that's not the case for BatchNorm or disabling track_running stats I still believe it's not ok to do this under the hood.

How about this:
- we add method `is_dp_compatible` to check 
- we do `validate` by default in `make_private`
- we add another method `fix_non_compliant_modules` so users could explicitly modify the model (and probably lol what we did).

I also see the confusion of the naming: `make_private` is indeed looks like we're about to do unspeakable things to the model, whereas the real intention was smth like `add_dp_enabling_wrappers`. Maybe rename this to `enable_dp` or `enable_privacy`?

PR #236, opened 2021-10-29
Autolint

alexandresablayrolles:
Add Github actions to run black and `flake8` and `autofix` (only works for `black`, not `flake8`)


karthikprasad:
Why do we need github actions for linting? Doesn't CircleCI already handle it for us?

PR #237, opened 2021-10-29
Add github action

alexandresablayrolles:
Add github action to auto run `black` and `flake8`, along with autofix (only works for `black`)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,28 @@
+name: Lint
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+      - experimental_v1.0
+
+jobs:
+  run-linters:
+    name: Run linters
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v2
+
+      - name: Run linters
+        uses: wearerequired/lint-action@v1
+        with:
+          auto_fix: true
+          black: true
+          flake8: true
+          flake8_args: "--config /Users/asablayrolles/code/utils/flake8_config.ini"
+


karthikprasad:
Duplicate of #236?

alexandresablayrolles:
> Duplicate of #236?

@karthikprasad: the nice thing about github actions is that there is an autofix for `black` that adds a commit fixing the lint on the PR.

romovpa:
Just curious, do we have plans to gradually move from CircleCI to Github Actions?

karthikprasad:
> Just curious, do we have plans to gradually move from CircleCI to Github Actions?

I don't think we have planned it yet, but since Pytorch is moving to GA, we might have to/benefit from moving. If we see benefits right sooner (looks like we are), might as well move sooner.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,28 @@
...
+        uses: actions/checkout@v2
+
+      - name: Run linters
+        uses: wearerequired/lint-action@v1
+        with:
+          auto_fix: true
+          black: true
+          flake8: true
+          flake8_args: "--config /Users/asablayrolles/code/utils/flake8_config.ini"
+
```
romovpa, line 28: How about running isort as well? 
There is [an action for it](https://github.com/isort/isort-action)

```yaml
- name: Run isort
  uses: isort/isort-action@master
  with:
    requirementsFiles: "requirements.txt dev_requirements.txt"
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,28 @@
+name: Lint
```
romovpa, line 1: I would mention in the name the fact that this is not just lint checks, but also auto-fixes.

ffuuugor:
On Github Actions vs CircleCI: will we be able to easily migrate GPU integration tests?

On this PR specifically: does this work for commits into main through phabricator? Awesome if yes!

romovpa:
Decided to close this at the Opacus core meeting.

PR #238, opened 2021-10-29
Collate packed sequences

alexandresablayrolles:
Support for tuples as well as single tensors

### DIFF
--- a/opacus/data_loader.py
+++ b/opacus/data_loader.py
@@ -1,6 +1,5 @@
 from typing import Optional, Sequence
 
-import collections
 import torch
 from opacus.utils.uniform_sampler import (
     DistributedUniformWithReplacementSampler,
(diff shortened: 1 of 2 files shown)


opacus/data_loader.py
```diff
@@ -10,21 +10,18 @@
 def wrap_collate_with_empty(
     collate_fn: Optional[_collate_fn_t], sample_empty_shapes: Sequence
 ):
-    # TODO: does it work with packed sequences?
     def collate(batch):
         if len(batch) > 0:
             return collate_fn(batch)
         else:
-            return [torch.zeros(x) for x in sample_empty_shapes]
+            if type(sample_empty_shapes) is collections.abc.Sequence:
```
karthikprasad, line 17: This should be isinstance() instead. 
`type(list()) is collections.abc.Sequence` will return False
`isinstance(list(), collections.abc.Sequence)` will return True

opacus/data_loader.py
```diff
@@ -10,21 +10,18 @@
...
+                return [torch.zeros(x) for x in sample_empty_shapes]
+            else:
+                return torch.zerso(sample_empty_shapes)
```
karthikprasad, line 20: typo: zeros

karthikprasad (changes requested):

ffuuugor:
Do we want to have a test for this, to make sure it works as expected?

alexandresablayrolles:
Turns out the original code was doing just fine...reverted my changes but still keeping tests in case it's useful.

ffuuugor (approved):
Yay, more tests!
Please fix flake8 errors before merging

PR #242, opened 2021-11-03
einsum problem with linear grad sample

pierrestock:
Make linear einsum compatible in the case of 3D activations (transformers use case). Maybe there is a way of getting rid of the `if/else` statement.

### DIFF
--- a/opacus/grad_sample/linear.py
+++ b/opacus/grad_sample/linear.py
@@ -21,7 +21,11 @@ def compute_linear_grad_sample(
         activations: Activations
         backprops: Backpropagations
     """
-    gs = torch.einsum("n...i,n...j->nij", backprops, activations)
+    # With activations with 3 dimensions (transformers) the original formula does not aggregate correctly
+    if activations.ndim == 2:
+        gs = torch.einsum("n...i,n...j->nij", backprops, activations)
+    else:
+        gs =  torch.einsum("nki,nkj->nij", backprops, activations)
     ret = {layer.weight: gs}
     if layer.bias is not None:
         ret[layer.bias] = torch.einsum("n...k->nk", backprops)
--- a/opacus/grad_sample/linear.py
+++ b/opacus/grad_sample/linear.py
@@ -21,7 +21,7 @@ def compute_linear_grad_sample(
         activations: Activations
         backprops: Backpropagations
     """
-    gs =  torch.einsum("nki,nkj->nij", backprops, activations)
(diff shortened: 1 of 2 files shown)


opacus/grad_sample/linear.py
```diff
@@ -21,7 +21,11 @@ def compute_linear_grad_sample(
         activations: Activations
         backprops: Backpropagations
     """
-    gs = torch.einsum("n...i,n...j->nij", backprops, activations)
+    # With activations with 3 dimensions (transformers) the original formula does not aggregate correctly
+    if activations.ndim == 2:
+        gs = torch.einsum("n...i,n...j->nij", backprops, activations)
+    else:
```
karthikprasad, line 27: Noob question: I thought ellipses in einsum handles all dimensions, and by extension even 3D. Do you know why it didn't work?

pierrestock: After careful investigation today, it appears that its linked to a previous version fo PyTorch that we used in our environment (specifically `1.7.1`). The bug appears to be solved for `1.9` since we could not reproduce for this version.

My suggestion would then be either (1) not to merge the PR but update the requirements for PyTorch or (2) merge the PR so that we have backward compatibility with older PyTorch versions.

image

ffuuugor: Updating PyTorch requirements past 1.8 is problematic, because we depend on csprng, which doesn't support 1.9 and beyond.
Is the bug still reproducable in  1.8?

pierrestock: Good news, bug disappears for 1.8.0. Let's update the requirements then?

image

karthikprasad: > Let's update the requirements then?

+1

romovpa: @ffuuugor Are you sure csprng doesn't support torch-1.9 (in it's current state). Did you see [my comment](https://github.com/pytorch/csprng/issues/130)? Just checking. I didn't test it deeply.

karthikprasad (approved):

opacus/grad_sample/linear.py
ffuuugor: You're probably right - but I haven't looked too deeply either.
I was only relying on the fact that PyPI release depend on 1.8 and their documentation explicitly lists all pyTorch versions they support, notably lacking 1.9 and beyond.
Is the fix as simple as installing from main branch instead of the latest release?

Issue #147, opened 2021-03-05
PyTorch Lightning issues with gradients not being detected properly


JMMarchant:
## 🐛 Bug

When trying to use PyTorch Lightning to create models, the `PrivacyEngine` fails due to `AttributeError: The following layers do not have gradients` despite the layers being included, implicitly, in the backward pass.

## Please reproduce using our [template Colab](https://colab.research.google.com/drive/1pCyN3_gN3WyhyQt8--eBPN5zVf2XZxiN?usp=sharing) and post here the link
https://colab.research.google.com/drive/1dLTMS7wZFD-43RNm1vdfW2umFkSTzfVB?usp=sharing

## Expected behavior

The backward pass used by PyTorch Lightning should work with the `opacus` `PrivacyEngine` hooks in the optimizer.

## Additional context

I have found this bug occurs across multiple environments (Colab, macOS, Ubuntu) and multiple PyTorch versions (1.6, 1.7.1)

sayanghosh:
@JMMarchant Thanks for filing this bug report, I believe it is related to the earlier issue https://discuss.pytorch.org/t/pytorch-lightning-support/113507.  We will be looking into Lightning integration with Opacus using this bug  you filed.

PR #244, opened 2021-11-04
PyTorch Lightning Demo (old API)

romovpa:
## Opacus + Lightning

[PyTorch Lightning](https://www.pytorchlightning.ai/) is one of the most used frameworks for PyTorch. It helps to remove boilerplate code, improve readability and reproducibility. Lightning does this by restructuring the code (and not changing), uses core PyTorch objects without wrapping them.

We want:
1. Provide PL users with a way to use Opacus
2. Adapt our API to add privacy learning with minumin changes (ideally with no changes) to LightningModule

## Proposed Changes

**Example** `mnist_lightning.py`. 
Demonstration of how to add privacy-preserving learning to PL-based code. This is a clone of `mnist.py` written with PL. The code is self-descriptive and uses [best practices](https://pytorch-lightning.readthedocs.io/en/latest/starter/style_guide.html).

**Fixed** important bug in DPOptimizer causing #147. 
The fix makes the example workable. Optimizer closure was used incorrectly in `dp_step()`: the model is not reevaluated before making a DP-step, as the result we don't have grad_samples to make a step. I believe this can affect training not only in the Lightning scenario.

_This is for the main branch, not v1.0._ 
Merging this change with v1.0 is straightforward, but we need to disassemble `make_private()` into parts. The code won't be the most elegant, but this approach is transparent and easy to debug/undestand.

## New API and Full Integration with Lightning

Creating a seamless integration is a challenging task:
- Opacus requires dependencies between model, optimizer and dataloader. Whereas Lightning forces separation between them.
- From Ligthning's perspective, Opacus implements a custom [`pl.Trainer`](https://pytorch-lightning.readthedocs.io/en/latest/common/trainer.html) flow. `Trainer` is not designed to be extended through inheritance. 
  - A custom Trainer could be a straightforward solution for user, but it works only in short-term and very hard to maintain. 
- `Trainer` promotes using [callbacks](https://pytorch-lightning.readthedocs.io/en/latest/extensions/callbacks.html) and [custom loops](https://pytorch-lightning.readthedocs.io/en/latest/extensions/loops.html) for extension. This seems to me the only viable way to integrate. In this case `Trainer` will need to either 
  1. mutate modules (attach and detach privacy-related stuff)
  2. check and accept only valid modules, check that LightningModule and DataModule is properly tied, throw exception otherwise

I will follow up with some options.

## Checklist

### DIFF
--- a/opacus/privacy_engine.py
+++ b/opacus/privacy_engine.py
@@ -352,6 +352,8 @@ class PrivacyEngine:
                 self.privacy_engine.steps += 1
                 return self.original_step(closure)
             else:
+                # Before running privacy_engine.step() we need to reevaluate the model and run backward() to compute
+                # grad_sample. In case these steps are provided in a closure, we need to run it.
                 loss = None
                 if closure is not None:
                     with torch.enable_grad():
--- a/opacus/privacy_engine.py
+++ b/opacus/privacy_engine.py
@@ -350,15 +350,22 @@ class PrivacyEngine:
             if hasattr(self.privacy_engine.module, "ddp_hooks"):
                 # We just update the accountant
                 self.privacy_engine.steps += 1
-
+                return self.original_step(closure)
             else:
+                loss = None
+                if closure is not None:
+                    with torch.enable_grad():
+                        loss = closure()
+
                 self.privacy_engine.step(is_empty)
                 if isinstance(
                     self.privacy_engine.module._module,
                     DifferentiallyPrivateDistributedDataParallel,
                 ):
                     average_gradients(self.privacy_engine.module)
-            self.original_step(closure)
+
+                self.original_step()
+                return loss
 
         def poisson_dp_step(self, closure=None):
             # Perform one step as usual
@@ -483,6 +490,7 @@ class PrivacyEngine:
 
         """
         self.steps += 1
+
         if not is_empty:
             self.clipper.clip_and_accumulate()
             clip_values, batch_size = self.clipper.pre_step()
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -142,6 +142,28 @@ commands:
       - store_artifacts:
           path: runs/mnist/test-reports
 
+  mnist_lightning_integration_test:
+    description: "Runs MNIST-Lightning example end to end"
+    parameters:
+      device:
+        default: "cpu"
+        type: string
+    steps:
+      - run:
+          name: MNIST-Lightning example
+          command: |
+            mkdir -p runs/mnist/data
+            mkdir -p runs/mnist/test-reports
+            echo "Using $(python -V) ($(which python))"
+            echo "Using $(pip -V) ($(which pip))"
+            python examples/mnist_lightning.py fit --trainer.accelerator <<parameters.device>> --model.lr 0.25 --model.sigma 0.7 --model.max_per_sample_grad_norm 1.5 --model.sample_rate 0.004 --trainer.max_epochs 1 --data.data_dir runs/mnist/data --data.sample_rate 0.004
+            python -c "import torch; exit(0)"
+          when: always
+      - store_test_results:
+          path: runs/mnist-lightning/test-reports
+      - store_artifacts:
+          path: runs/mnist-lightning/test-reports
+
   cifar10_integration_test:
     description: "Runs CIFAR10 example end to end"
     parameters:
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -15,3 +15,4 @@ tensorboard
 datasets
 transformers
 scikit-learn
+pytorch-lightning[extra]
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -15,4 +15,5 @@ tensorboard
 datasets
 transformers
 scikit-learn
-pytorch-lightning[extra]
+pytorch-lightning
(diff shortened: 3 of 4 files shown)


alexandresablayrolles:
LGTM, thanks @romovpa for the diff!

romovpa:
[Implemented a prototype](https://github.com/romovpa/opacus/pull/5) of the better API-level integration with PL.

A PyTorch-Lightning user does:
```python
# Modeling as usual. No privacy!
class LitModel(LightningModule):
    def __init__(self):
        super().__init__()
        ...

    def forward(self):
        z = torch.rand(1, 3)
        image = self.decoder(z)
        image = image.view(1, 1, 28, 28)
        return image


def main():
    # Initialize model and data loader. No privacy here.
    model = LitModel()
    data = MNISTDataModule()

    # Adding some differential privacy: creating an engine, wrapping the DataModule
    privacy_engine = LightningPrivacyEngine(
        delta=1e-5,
        sample_rate=0.001,
        # etc
    )
    privacy_data = privacy_engine.wrap_datamodule(data)

    # Training
    trainer = pl.Trainer(
        max_epochs=2,
        enable_model_summary=False,
        callbacks=[privacy_engine],  # adding callback provided with the PrivacyEngine
    )
    trainer.fit(model, privacy_data)  # using a valid data module

    # This will crash with the message: 
    # Either remove PrivacyEngine callback or use certified data module
    trainer.fit(model, data)

    # Testing is as usual, no privacy here
    trainer.test(model, data)
    trainer.test(model, privacy_data)  # this also works

```

cc @ffuuugor @alexandresablayrolles

romovpa:
Anyway, I suggest to merge & close this PR to guarantee that we have at least something for PL users, if they want to use Opacus. As a follow up I will try to suggest an API in `experimental_v1.0` branch.

To implement the idea above I will probably need to [redesign `PrivacyEngine` as a factory](https://github.com/pytorch/opacus/pull/226#discussion_r733789307), if there is no objection. This approach divides hyperparameters (settings) and the arguments to which DP is applied (model, data, opt). I need this separation for a clean integration with PL.

ffuuugor:
Thanks!
Agree, let's merge it to main ASAP - I see no reason not to.
If you import the diff, we'll accept 

re: factory redesign 
I was going to implement the proposal almost exactly like stated in the comment you linked.
One difference - because of unlucky timing of PyTorch DevDay recordings, we've already committed to a certain API, so we probably want to also support simple constructor object creation too.
In terms of which parameters go where, I feel strongly (and also we've committed to in the presentation) that noise_multiplier and max_grad_norm belong in `make_private`.
`poisson_sampling` should probably be moved to constructor (and therefore factory), as it defines the behaviour strategy, not just parameters. 

I was going to implement that over the next couple days (certainly before Tuesday) - does this timeline work for you? If not, feel free to do the draft implementation and we can converge later

ffuuugor (approved):

opacus/privacy_engine.py
```diff
@@ -350,15 +350,22 @@ def dp_step(self, closure=None, is_empty=False):
...
                 # We just update the accountant
                 self.privacy_engine.steps += 1
-
+                return self.original_step(closure)
             else:
+                loss = None
+                if closure is not None:
+                    with torch.enable_grad():
+                        loss = closure()
+
```
karthikprasad, line 359: noob: Why isn't this necessary when module has `ddp_hooks`? Why can't this be common for both if and else blocks?

romovpa: Good question. It is worth explaining. 

First, wth is the closure for. From [PyTorch docs](https://pytorch.org/docs/stable/optim.html):

> Some optimization algorithms such as Conjugate Gradient and LBFGS need to reevaluate the function multiple times, so you have to pass in a closure that allows them to recompute your model. The closure should clear the gradients, compute the loss, and return it.
> 
> Example:
> ```python
>     def closure():
>         optimizer.zero_grad()
>         output = model(input)
>         loss = loss_fn(output, target)
>         loss.backward()
>         return loss
> ```

Before clipping and adding noise in `self.privacy_engine.step(is_empty)` we need to compute the model. Otherwise there is nothing to apply clipping/noise to. 

In case of `ddp_hooks` we just call `self.original_step(closure)` and pass the closure to call it at the original oprimiser's discretion. As the comment suggests, the clipping and noising are performed by the hooks at the end of the backward pass.

opacus/privacy_engine.py
```diff
@@ -350,15 +350,22 @@ def dp_step(self, closure=None, is_empty=False):
...
                 self.privacy_engine.step(is_empty)
                 if isinstance(
                     self.privacy_engine.module._module,
                     DifferentiallyPrivateDistributedDataParallel,
                 ):
                     average_gradients(self.privacy_engine.module)
-            self.original_step(closure)
+
+                self.original_step()
```
romovpa, line 367: Hmm, I'm not sure if I should pass the closure to the original_step. 
- If I do, it probably calls it again and we run the backward pass twice. 
- If I do NOT, the optimizers requiring several model reevaluations must not work. Although don't know how that case affects grad sampling and accounting.

@alexandresablayrolles maybe you know the answer?

opacus/privacy_engine.py
karthikprasad: I see. Thanks @romovpa . Can you perhaps add the above explanation as a comment to the code. (Unless, it is obvious for everyone.)

PR #247, opened 2021-11-06
Merge from main

ffuuugor:
Bring latest changes from the main branch.
The most important aspect - it includes fixes for integration tests, so now they should be passing in the experimental branch too.


opacus/optimizers/optimizer.py
```diff
@@ -167,13 +163,12 @@ def pre_step(self) -> Optional[float]:
...
-        
-        return self.optimizer.step(closure)
-
-
-    def pre_step(self):
-        self.clip_and_accumulate()
+        if self.pre_step():
+            return self.optimizer.step(closure)
+        else:
+            return None
```
romovpa, line 172: There is probably a bug here. 

We do `self.clip_and_accumulate()` before the model is reevaluated by calling the closure, we should not have appropriate sample_grads to work with. This breaks optimization loop when using Lightning's [automatic optimization](https://pytorch-lightning.readthedocs.io/en/latest/common/optimizers.html#automatic-optimization). 

This may work incorrectly with the manual optimization loop in our examples: we must be using grad_samples from the previous iteration, not the current one. I'm not sure though.

I tried to fix this for the old PrivacyEngine [here](https://github.com/pytorch/opacus/pull/244/files#r744205844).

opacus/optimizers/perlayeroptimizer.py
```diff
@@ -22,10 +21,16 @@ def __init__(
...
+            optimizer,
+            noise_multiplier=noise_multiplier,
+            max_grad_norm=max_grad_norm,
+            expected_batch_size=expected_batch_size,
+            loss_reduction=loss_reduction,
+        )
 
     def attach(self, optimizer):
-       self.optimizer = optimizer
+        self.optimizer = optimizer
```
romovpa, line 33: Not an issue, but do I understand correctly that if we have `optimizer = DPPerLayerOptimizer()` then the original optimizer is `optimizer.optimizer.optimizer`? 😆

romovpa: I wanted to ask though, is it a good pattern to inherit a class and store the parent's instance in the child's fields? Probably it is. Although I have a feeling that it may be possible to simplify by share fields with the parent and use super() calls.

opacus/tests/scheduler_test.py
```diff
@@ -46,7 +46,9 @@ def test_step_scheduler(self):
         self.assertEqual(self.optimizer.noise_multiplier, gamma ** 2)
 
     def test_lambda_scheduler(self):
-        noise_lambda = lambda epoch: (1 - epoch / 10)
+        def noise_lambda(epoch):
+            return 1 - epoch / 10
```
romovpa, line 50: Btw, why not use lambda syntax here?

I would even put it right in place where it is used (sorry, I'm allergic to spaghetti code):
```python
scheduler = LambdaNoise(
    self.optimizer,
    noise_lambda=lambda epoch: 1 - epoch / 10,
)
```

opacus/validators/__init__.py
```diff
@@ -1,13 +1,13 @@
...
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 
-from .batch_norm import validate, fix  # noqa
-from .ddp import validate, fix  # noqa
-from .instance_norm import validate, fix  # noqa
-from .lstm import validate, fix  # noqa
+from .batch_norm import fix, validate  # noqa
+from .ddp import fix, validate  # noqa
+from .instance_norm import fix, validate  # noqa
+from .lstm import fix, validate  # noqa
```
romovpa, line 7: What was the reason of doing this?
I would wrap this in a function to make it self-explanatory. Otherwise isort will mix everything together.

```python
from .module_validator import ModuleValidator
from .utils import register_module_validator, register_module_fixer

def check_validator_submodules():
    from .batch_norm import fix, validate  # noqa
    from .ddp import fix, validate  # noqa
    from .instance_norm import fix, validate  # noqa
    from .lstm import fix, validate  # noqa
    from .multihead_attention import fix, validate  # noqa

check_validator_submodules()

__all__ = [
    "ModuleValidator",
    "register_module_validator",
    "register_module_fixer",
]

```

examples/pysift.py
```diff
@@ -0,0 +1,254 @@
+import numpy as np
```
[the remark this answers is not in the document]
ffuuugor: Whoops, that wasn't supposed to be here.
I was just playing around with pysift. Need to be careful with this `git add examples`

opacus/optimizers/perlayeroptimizer.py
ffuuugor: Way ahead of you: https://github.com/pytorch/opacus/blob/beeac9b50d59ea6304b7fb5be62d3d08f6e5ad99/opacus/privacy_engine.py#L232

I mean, yeah, we can do better at naming here.

But on a serious note, I'd love to have this as a proper inheritance, but I couldn't find a way.
Original optimizer can be of any `Optimizer` subclass, which makes it hard to handle this through inheritance

opacus/tests/scheduler_test.py
ffuuugor: I was surprised too
https://www.flake8rules.com/rules/E731.html

opacus/validators/__init__.py
ffuuugor: @karthikprasad wdyt?

I think the purpose of this is to make sure `register_validator` methods are called - and I'm not sure what's the best way to make it obvious for the reader here

opacus/optimizers/optimizer.py
ffuuugor: ouch
tbh, it is true that `closure` was an afterthought here. I wasn't familiar with the applications you've listed in your PR.

would simply calling `closure()` before `clip_and_accumulate()` do the trick?

opacus/validators/__init__.py
karthikprasad: > I think the purpose of this is to make sure register_validator methods are called 

That is correct. I don't know if there is an accepted convention.

> and I'm not sure what's the best way to make it obvious for the reader here

Perhaps add Igor's explanation as a comment on the top of the file explaining the need for such imports?

> I would wrap this in a function to make it self-explanatory.

Wrapping in a function is also fine, but I don't think it is self-explanatory. A comment might still be necessary, in which case, we can do away with the function and simply use a comment? WDYT?

> Otherwise isort will mix everything together.

What is the issue here? The order of import doesn't matter as long as all are imported.

opacus/tests/dp_layers/dp_rnn_test.py
```diff
@@ -43,12 +44,15 @@ class DPLSTM_test(DPModules_test):
         batch_first=st.booleans(),
         zero_init=st.booleans(),
         packed_input_flag=st.integers(
-            0, 2
-        ),  # 0 indicates no packed sequence input, 1 indicates packed sequence input in sorted order, 2 indicates packed sequence input in unsorted order
+            0,  # no packed sequence input
+            # 1,  # packed sequence input in sorted order
```
karthikprasad, line 48: noob: why is this not tested?

opacus/layers/param_rename.py
```diff
@@ -18,16 +18,28 @@ def filter_out_old_keys(self, state_dict, prefix, local_metadata):
     return new_state_dict
 
 
-class ParamRenamedModule(nn.Module):
+class RenameParamsMixin:
```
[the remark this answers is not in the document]
ffuuugor: Was changed in #222 
@romovpa can provide more context, but I think this was a good change: mixins give you more flexibility compared to classic inheritance

opacus/tests/dp_layers/dp_rnn_test.py
ffuuugor: oh
it it considered, but the comment is extremely misleading. The change is again coming from #222 

`st.integers(0,2)` actually sets a range -> it iterates over [0,1,2]
The intention of this comment is to provide an explanation for each input flag value, while integrated into the code. But I think it failed to do so, as it looks obviously as if we're skipping `1`.
I'll revert to old comment

opacus/validators/__init__.py
romovpa: A comment is a must for the future contributors to understand why it is necessary to import submodules. Igor's explanation is good enough. Thanks, now I know this too :)

> What is the issue here? The order of import doesn't matter as long as all are imported.

The order doesn't matter for the lib to work, but is does matter for the human to read. The way the code structured is like a docstring.

> we can do away with the function and simply use a comment? 

Agree, but now all imports are mixed up, there is no good place to put explanation, look:

```python
from .lstm import fix, validate  # noqa
from .module_validator import ModuleValidator
from .multihead_attention import fix, validate  # noqa
from .utils import register_module_fixer, register_module_validator
```

If all `register_validator`-related imports are in one place it is more clear for the contributor that they need to put the module here after adding a new validator.

opacus/optimizers/optimizer.py
romovpa: Adding this in the beginning of `step()` should partially solve the problem:

```python
                loss = None
                if closure is not None:
                    with torch.enable_grad():
                        loss = closure()
```

In the end we need to return `loss` instead of `self.optimizer.step(closure)` output. 

However, [I don't know what happens](https://github.com/pytorch/opacus/pull/244/files#r744265353) in case the optimizer makes several backward passes during one step. It is not clear to me if it can break DP accounting. 

I need to investigate this. For this PR I would either merge it as is or add my fix, but in any case leave a comment / TODO.

opacus/layers/param_rename.py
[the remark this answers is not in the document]
romovpa: > is this intentionally no longer a subclass of nn.Module?

Right, now it's intentionally a [trait](https://en.wikipedia.org/wiki/Mixin). It's a more suitable pattern here. It allows you to add params rename without breaking or cluttering the existing subclass structure of your Module.

opacus/tests/dp_layers/dp_rnn_test.py
romovpa: My bad, misunderstood the meaning.

opacus/tests/scheduler_test.py
romovpa: Right, but if we use lambda inline to pass the argument (as in my example), it's not a violation of E731.

opacus/layers/param_rename.py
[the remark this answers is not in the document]
karthikprasad: Thanks. This is indeed more flexible.

opacus/validators/__init__.py
karthikprasad: > Agree, but now all imports are mixed up, there is no good place to put explanation, look:

Ouch! Ugly!

> If all register_validator-related imports are in one place it is more clear for the contributor that they need to put the module here after adding a new validator.

+3000

karthikprasad (approved):
Thanks @ffuuugor . This is a massive merge.
Accepting since most unit tests and all integration tests are passing. Please take a loot at multi_gpu unit test as well before merging.

ffuuugor:
Thanks for accepting @karthikprasad 
I was going to do multi_gpu test in a separate PR - it can take quite a while to debug, and in the meantime I want to unblock all the other PRs to be able to use CircleCI for validation

Issue #225, opened 2021-10-14
Attached optimizer should not touch parameters that weren't passed to it


ndrmnl:
## 🐛 Bug

The `step` done by the `PrivacyEngine` considers all the parameters that have `requires_grad = True`, whether or not they have been passed to the attached optimizer.

An easy fix is to go through `optimizer.param_groups` rather than through the `self.module.parameters` in L501 of the `PrivacyEngine`.

## To Reproduce

A `step` on the `dp_optimizer` below will also update layer 1.

```
model = torch.nn.Sequential(torch.nn.Linear(5, 10), torch.nn.Linear(10, 1))

dp_params = [p[1] for p in model.named_parameters() if p[0].startswith('0')]
dp_optimizer = torch.optim.SGD(dp_params, lr=0.01)
privacy_engine.attach(dp_optimizer)
```

## Expected behavior

The optimizer shouldn't touch parameters that were not passed to it.

ffuuugor:
Hi @ndrmnl, and thanks for pointing out the issue - the expected behaviour you described is indeed what should happen.

We're currently actively working on a new release of Opacus with a major revamp of the API to simplify the code (see branch `experimental_v1.0`). The issue described here is already fixed in the new version and will be available once we ship the changes (plan is mid-November). 

As for the short-term, I don't think we'll be able to prioritise that - the long-term fix is coming relatively soon, and I don't think this is going to be a one-line change (in addition to L501 you'd also need to adjust the clipping and how you calculate sample grad norm)

Thanks again for opening this issue. Let's leave it open and close it once we land the changes and verify they resolve the problem

PR #248, opened 2021-11-07
add API to get compatible modules

karthikprasad:
## Types of changes

From @ffuuugor:
> There could be an issue with doing model fixing and module replacement in make_private.
Imagine scenario (not a hypothetical, I'm fixing dcgan example now):
> 1) you create a model and an optimizer. And you pass model.parameters() to the optimizer, because that's how you initialise optimizer
> 2) you then pass aforementioned model and optimizer to make_private() with try_fix_incompatible=True. It does clone_module for the root and a bunch of children and returns the result.
> 3) Now we have a problem: optimizer is pointing to the parameters of the old model before fixing, but gradients are computed on the parameters of the new model, after fixing

We can't simply clone the optimizer and supply new model's parameters either (see #225)

unittest passes on top of ffuuugor_experimental_merge

## Checklist


romovpa:
@karthikprasad Check plz if I understood the solution correctly. 

If the model is not valid, we suggest user to apply the transform and fix it explicitly:
```
model = models.resnet18()
fixed_model = privacy_engine.get_compatible_module(model)
```

After this, `model` and `fixed_model` are two independent models with their own physical parameters. As a common case we should use the latter for further computations.

opacus/privacy_engine.py
```diff
@@ -207,6 +203,9 @@ def _prepare_model(
         try_fix_incompatible_modules: bool = False,
```
ffuuugor, line 203: do we want to remove `try_fix_incompatible_modules` from here as well?

opacus/tests/privacy_engine_test.py
```diff
@@ -102,25 +102,25 @@ def _init_private_training(
         noise_multiplier: float = 1.0,
         max_grad_norm: float = 1.0,
         poisson_sampling: bool = True,
-        try_fix_incompatible_modules: bool = False,
     ):
+        privacy_engine = PrivacyEngine(secure_mode=secure_mode)
+
         model = SampleConvNet()
+        model = privacy_engine.get_compatible_module(model)
```
ffuuugor, line 109: I'd rather have a test both both occasions: where we call `get_compatible_module` and where we don't.
Both should work (obvs provided the model is compatible in the first place). 

The original bug was spotted by dumb luck: `dcgan` example just turned out to be the only place where we used fixing behaviour, but it wasn't designed for it.
Let's make sure we have dedicated test for this

ffuuugor (changes requested):
Thanks for a prompt and solid fix @karthikprasad  !
(No, really, Karthik submitted the diff less than 30 minutes after a bug was discovered. We still move fast sometimes)

Few inline comments

opacus/privacy_engine.py
```diff
@@ -56,6 +56,15 @@ def validate(
         """
         ModuleValidator.validate(module, raise_if_error=True)
 
+    def get_compatible_module(self, module: nn.Module) -> nn.Module:
```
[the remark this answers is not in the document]
karthikprasad: >We can make it a @classmethod.

I don't see why not. Makes more sense since it doesn't depend on any instance attributes.

>what's the downside of asking people to call ModelValidator.fix() directly?

1. Users having to import `ModuleValidator` in their training file (in addition to PrivacyEngine).
2. We lose the [single-window system ](https://www.wikiwand.com/en/Single-window_system) for accessing Opacus internals.

karthikprasad:
> @karthikprasad Check plz if I understood the solution correctly.
> 
> If the model is not valid, we suggest user to apply the transform and fix it explicitly:
> 
> ```
> model = models.resnet18()
> fixed_model = privacy_engine.get_compatible_module(model)
> ```
> 
> After this, `model` and `fixed_model` are two independent models with their own physical parameters. As a common case we should use the latter for further computations.

@romovpa That is correct!

opacus/privacy_engine.py
karthikprasad: I kept it as is since, technically the user can call `privacy_engine._prepare_model()` with the flag. However, the method is meant to be private, so it doesn't make sense anyway. I'll remove it

opacus/privacy_engine.py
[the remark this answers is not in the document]
ffuuugor: >Users having to import ModuleValidator in their training file

That make sense. I don't have a strong preference, and this argument alone is enough to justify this decision I think.

> We lose the single-window system for accessing Opacus internals.

As a more general point, I'd love to have a discussion about this at some point. I would argue that having a single-window system is not the most fitting design approach, despite opacus historically being this way.

ffuuugor (approved):
LGTM, Thank you!

JFYI, I've merged the main branch, so now the only two tests that are expected to fail are multi_gpu and linter on vision_benchmark.

PR #250, opened 2021-11-08
[WIP] PyTorch Lightning Demo for API v1.0

romovpa:
Example demonstrating how Opacus v1.0 API can be used with PyTorch Lightning. This is a follow-up for #244. 

## Changes

Added `mnist_lightning.py` with the classifier allowing training with DP. 

TODO:
- [x] Fix the problem with `forbid_accumulation_hook`
- [ ] Fix attaching GradSampleModule twice when using LightningCLI
- [ ] Poisson sampling: check the data loader is consistend with the optimizer
- [ ] Correctly handle DDP (raise NotImplementedError, for example)
- [ ] Add a guide on adding DP to Lightning code in the docstring

## What about Full Integration with Lightning?

There are two levels of integration:
1. Provide PL users with a straightforward and clean way to add DP training by **modifying** their `LightningModule`.
2. Provide PL users with a **seamless** solution "turn on privacy" working on the Trainer level and requiring no changes inside `LightningModule`. 

Having a **seamless** integration is desirable as this 1) lowers the entry threshold to a minimum (thus making privacy-in-ML a commodity not magic); 2) preserves the separation of logical parts promoted by PL.

Having a **modify-module** integration is a MUST. The lack of it will force PL users to rewrite their code to pure torch, thus discouraging usage of privacy-preserving training.

## How do we build a Full Integration?

Below are my thoughts after exploring ways of implementing integration with PL. This doesn't pretend to be a detailed analysis, just thoughts. 

Prototypes for the old API:

1. Modify-module #244
    - ✅ Clean & straightforward
    - Main methods of LitghtningModule remains unchanged (forward, configure_optimizers, training/test_step)
    - DP is enabled by attaching PrivacyEngine to the optimizer in `on_train_start` method
2. Seamless [romovpa/#5](https://github.com/romovpa/opacus/pull/5)
    - ✅ Clean and straightforward
    - No changes in LightningModule is required
    - DP is enabled by adding a special callback to the `pl.Trainer`
    - If LightningDataModule is used, we enable poisson sampling by wrapping the data module.

The anticipated v1.0 API moves from attach/inject architecture to wrap/extend. At least on first approach, this creates an obstacle for integrating with PL:

1. Modify-module #250 (this PR)
     - ✅ Straightforward & clean enough.
     - DP is enabled by calling `make_private()` in `configure_optimizers()` to convert the optimizer to DP version
     - 👌 No direct way to access training data in LightningModule, have to use private fields (not good). Solving [this issue](https://github.com/PyTorchLightning/pytorch-lightning/issues/10430) will help. Folks from the Lightning team assured me that the dataloader will be officially available in `trainer` eventually. In this case it is safe to proceed with the hack.
2. Subclass-module (next PR)
    - ✅ Straightforward & clean
    - DP is enabled via inheriting from `DPLightningModule` which hides pre-defined parameters and necessary wrappings from the users.
3. Seamless (future)
    - 🤔  Requires integrating **INTO** Lightning
        - DP will be enabled via Trainer settings and implemented directly in `trainer.fit(model, data)`.
        - 💪 Opacus must be a dependency of Lightning which is a big step. We need to get community support. We must ensure stability of the main branch.
        - Can we use standard extension mechanisms? No.
            - [Callback](https://pytorch-lightning.readthedocs.io/en/latest/extensions/callbacks.html): we cannot wrap model/optimizer. Although we can use callback to attach/add hooks (see [this](https://github.com/romovpa/opacus/pull/5)).
            - [TrainingTypePlugin](https://pytorch-lightning.readthedocs.io/en/latest/extensions/plugins.html#training-type-plugins): intended to customize training running in parallel on many machines.

## Other questions 

- ‼️ Not clear how to combine Opacus' distributed modules and PL's. PL is designed to separate the model from the training implementation, so DDP is deeply under the hood.

## Related

- Forum discussion: [PyTorch Lightning Support](https://discuss.pytorch.org/t/pytorch-lightning-support/113507/5)
- Lightning issue: [Add a callback to support ocapus](https://github.com/PyTorchLightning/pytorch-lightning/issues/3305)
- Lightning issue: [Unable to access datalaoders within configure_optimizers](https://github.com/PyTorchLightning/pytorch-lightning/issues/10430)


romovpa:
Met with @four4fish from the Lightning team (thank you, Siyu!). 

- Got assurance that `dataloader` will be available in `trainer`. This means the hack we currently use is ok, later we change it to a better option.
- Found out that the existing extension methods (Plugins, Callbacks, etc) cannot suit our needs.
- Came up with an alternative way to add DP in LightningModule using a different base class. I will implement this in the next PR.
- The best way to implement seamless integration is to make Opacus a dependency of PL and incorporate necessary modifications right into `trainer.fit()`. For this we need support of the Lightning community.

ffuuugor:
> The best way to implement seamless integration is to make Opacus a dependency of PL and incorporate necessary modifications right into trainer.fit(). For this we need support of the Lightning community.

Do you think this is the right move for them? As much as I'd love to have clean integration, I understand that taking such a dependency is a big step, especially for something as experimental as DP-SGD

opacus/privacy_engine.py
```diff
@@ -26,11 +26,12 @@ def forbid_accumulation_hook(module: nn.Module, _):
...
         if hasattr(p, "grad_sample"):
+            pass
             # TODO: this correspond to either not calling optimizer.step()
             # or not calling zero_grad(). Customize message
-            raise ValueError(
-                "Poisson sampling is not compatible with grad accumulation"
-            )
+            # raise ValueError(
+            #     "Poisson sampling is not compatible with grad accumulation"
+            # )
```
ffuuugor, line 34: 🦖

examples/mnist_lightning.py
```diff
@@ -0,0 +1,169 @@
...
+    def configure_optimizers(self):
+        optimizer = optim.SGD(self.parameters(), lr=self.lr, momentum=0)
+
+        if self.enable_dp:
+            # dirty introspection of the trainer instance to get the training data
+            data_loader = (
+                self.trainer._data_connector._train_dataloader_source.dataloader()
+            )
+            # transform (model, optimizer, dataloader) to DP-versions
+            model, optimizer, dataloader = self.privacy_engine.make_private(
```
ffuuugor, line 95: Just trying to understand what's happening here:
data_loader on line 91 is already a DPDataLoader, right?
So we don't care that data_loader returned from `make_private` is discarded

opacus/privacy_engine.py
ffuuugor, line 34: 🦖

ffuuugor (commented):
LGTM!
I see this is still in draft state - what's missing for the merging?

opacus/privacy_engine.py
romovpa: Not ready, need to fix this. If uncommented, this exception gets raised for some reason.

examples/mnist_lightning.py
romovpa: Following the Lightning philosophy, the module should be independent from data (we may pass different data to `trainer.fit(module, data)`). This means here we shouldn't expect wrapped data_loader automatically

Here `make_private` does nothing with the dataloader because we pass `DPLightningDataModule` to `trainer.fit`. 

But it we do not pass wrapped model, this works incorrectly (we will use the original dataloader). I'm going to add a check here and raise if the data is not wrapped and we do poisson sampling, or if we cannot get sample_rate.

romovpa:
> Do you think this is the right move for them? As much as I'd love to have clean integration, I understand that taking such a dependency is a big step, especially for something as experimental as DP-SGD

@ffuuugor I think it is. Siyu suggested this and I agree. 

You are right, this is a big step. As I commented above, not only do we need to have Opacus ready for being a dependency for Lightning, we also need to convince the community that this feature worth including it in the Lightning. I believe we should be able to accomplish this at some point in the future. I also believe this can be a good indicator for Opacus' maturity.

ffuuugor:
Interesting and promising direction! Thanks @romovpa

romovpa:
Moving to #298

Issue #229, opened 2021-10-19
Start using Doctests in Opacus


romovpa:
## 🎨 Enhancement

Start using [doctest](https://docs.python.org/3/library/doctest.html). Make existing examples proper doctests, remove  [`docstring_examples_test.py`](https://github.com/pytorch/opacus/blob/main/opacus/tests/docstring_examples_test.py).

For example, [this code](https://github.com/pytorch/opacus/blob/78e2f49c6498de206ca06c4da69e5e53763779f4/opacus/dp_model_inspector.py#L83) doesn't need [additional test](https://github.com/pytorch/opacus/blob/78e2f49c6498de206ca06c4da69e5e53763779f4/opacus/tests/docstring_examples_test.py#L37) as the example can be run automatically:
```python
def validate(self, model: nn.Module) -> bool:
    r"""
    Runs the validation on the model and all its submodules.
    ...
    Example:
        >>> inspector = DPModelInspector()
        >>> valid_model = nn.Linear(16, 32)
        >>> is_valid = inspector.validate(valid_model)
        >>> is_valid
        True
        >>> invalid_model = nn.BatchNorm1d(2)
        >>> is_valid = inspector.validate(invalid_model)
        # IncompatibleModuleException is thrown.
    """
```

## Motivation

- We don't need to sync examples with tests. (Get rid of _IMPORTANT: When changing this code you also need to update xxx_)
- Docstings naturally become more informative.
- Additional test coverage. Docstrings cover shallow API checks that are hard to cover with main tests. 

## What to do

- [ ] Integrate [doctest](https://docs.python.org/3/library/doctest.html) or an alternative with the CI
- [ ] Fix all doctests mentioned in [docstring_examples_test.py](https://github.com/pytorch/opacus/blob/main/opacus/tests/docstring_examples_test.py)
- [ ] (stretch) Identify other places where doctests would be useful

PR #252, opened 2021-11-10
Doctests: running examples from docstings automatically

romovpa:
## Doctests

Issue and motivation: #229. We are going to write plenty of tests. I want to encourage the team to use doctests whenever it's practical. By doing so we feed two birds with one seed: improve documentation and test coverage.

[Doctest](https://docs.pytest.org/en/6.2.x/doctest.html) are just docstrings run as unit tests with pytest. For example:
```python
class SequenceBias(nn.Module):
    r"""...

    Example:
        >>> m = SequenceBias(16)
        >>> input = torch.randn(20, 4, 16)
        >>> output = m(input)
        >>> print(output.size())
        torch.Size([21, 4, 16])
    """
```

Pytest:
```
$ pytest --doctest-modules -p conftest opacus/layers/dp_multihead_attention.py
=========================== test session starts ===========================
platform darwin -- Python 3.9.7, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /Users/romovpa/workspace/opacus
plugins: hypothesis-6.24.2
collected 1 item                                                          

opacus/layers/dp_multihead_attention.py .                           [100%]

============================ 1 passed in 0.01s ============================
```

## Changes

- Exising doctests are fixed.
- Removed `docstring_examples_test.py`. 
    - We don't need to synchronize tests with docstrings anymore
    - Disabled PrivacyEngine tests are moved to `privacy_engine.py` as TODO 
- Circle-CI command modified

All pytests passed (including doctests):
```
$ pytest --doctest-modules -p conftest opacus
```

## Further steps

- Write guidelines on how to use doctests efficiently
- Update CONTRIBUTING


ffuuugor (approved):
Great initiative, thanks for following through on your suggestion in the issues!

Just to clarify, what's the correct way of using code in the docstrings so it's picked up by `doctests`?

romovpa:
> Just to clarify, what's the correct way of using code in the docstrings so it's picked up by `doctests`?

Basically, just add examples in the docstrings. Pytest will detect them automatically and run as test.

Some things you need to know:

1. **Formatting**. The code should be formatted as if you were using REPL: 
    - `>>>` means start of an expression
    - `...` means the continuation of the expression
    - The text after the expression is a correct expected output
2. **Skip**. By default all the code in docstrings will be run. If you don't want to test expression, put a comment `# doctest: +SKIP` 
3. **Output comparison**. Copy the output exactly as it is in REPL.
    - You can put `...` in the docstring's output, this will match any substring. For example `tensor([...])` will match any inline tensor. This is handy when using randomized input.
    - You can add `# doctest: +NUMBER` on the expression, this will compare numbers automatically rounding them to the precision used in the docstring.
4. **Namespace**. 
    - If you want to use a module which is not imported outside of the docstring, import it in the docstring.
    - Some frequently used modules and classes are added automatically in the test namespace, such as `torch`, `nn`, `np`, `opacus`, etc. See `conftest.py`.
    - We can add toy classes to demonstrate functions on them. For example I created `MyCustomModule` as a real `nn.Module`, it is available to use in docstring tests.

karthikprasad (approved):
Great initiative indeed @romovpa ! Thanks a lot for this suggestion and the PR 😃 

> Some things you need to know:

Not as part of this PR, but it would be good to add your comment to CONTRIBUTING.md file as well for developers to follow.

romovpa:
@ffuuugor fyi, `test_noise_level` has failed (flaky)

ffuuugor:
> @ffuuugor fyi, test_noise_level has failed (flaky)

Thanks for the heads up. That's expected, I've addressed this in one of the open PRs:

https://github.com/pytorch/opacus/pull/254/files#diff-af4fd706ae53070877bde3d1dc350c8bbbdc51db6e88b4b07748dc3a983334deR45

PR #253, opened 2021-11-10
revive multi_gpu_gradcheck

ffuuugor:
Updated to the the new API, works as expected


opacus/tests/multigpu_gradcheck.py
```diff
@@ -68,36 +69,41 @@ def demo_basic(rank, weight, world_size, dp):
...
-    optimizer = optim.SGD(ddp_model.parameters(), lr=1)
+    privacy_engine = PrivacyEngine()
+
     if dp:
-        engine.attach(optimizer)
+        ddp_model, optimizer, data_loader = privacy_engine.make_private(
+            module=ddp_model,
+            optimizer=optimizer,
+            data_loader=data_loader,
+            noise_multiplier=0,
```
romovpa, line 92: Why zero noise?

opacus/tests/multigpu_gradcheck.py
```diff
@@ -68,36 +69,41 @@ def demo_basic(rank, weight, world_size, dp):
 
     # create model and move it to GPU with id rank
     model = ToyModel().to(rank)
+    optimizer = optim.SGD(model.parameters(), lr=1)
```
karthikprasad, line 72: Can't comment on unchanged lines!! Ugh! I hate github's restrictions.
nit: remove L38

karthikprasad: Also, the test won't run on win32 in its current form, right? due to L27?

karthikprasad (approved):
couple of nits. LGTM otherwise.

opacus/tests/multigpu_gradcheck.py
ffuuugor: good catch: shall I just delete win32 part? I don't think we intend this as a public tutorial, as I presume our test env will always stay unix

ffuuugor: line of reasoning: you need manual intervention for running this on windows (provide a filename) -> it's not worth having as a test

karthikprasad: > you need manual intervention for running this on windows (provide a filename)

Wow! If it can't be automated and not a tutorial, I think we should remove it. If that portion of the code never runs, we will never know if it is doing its job correctly or not.

Curious though, is there no way to run automated tests on windows?

ffuuugor: No first hand experience, but my impression from reading the docs is that you need to provide a filename for syncronization, and windows don't have a conventional `/tmp` folder

PR #254, opened 2021-11-10
PrivacyEngineFactory

ffuuugor:
This PR introduces `PrivacyEngineFactory` as a main entry point for non-standard (i.e. non-flat flipping, non-poisson sampling, non-rdp accounting) PrivacyEngines. 

## Example:
```
from opacus import PrivacyEngineFactory

privacy_engine = PrivacyEngineFactory.get(clipping = "per_layer", poisson = False)
model, optimizer, data_loader = privacy_engine.make_private(...)
```

We also make the "old" (one in the DevDay presentation) available.

### Mixins as a mechanism to express different strategies in PrivacyEngine
Example of class definion:
```
class PrivacyEngine(
    PoissonDataLoaderMixin, RDPAccontantMixin, PrivacyEngineFlatClippingBase
):
    pass
```

We could've kept pure inheritance, but with the prospect of more accountants it'll soon become too verbose.
For now this works well, but only as long as different Mixins (e.g. accountant and data_loader) don't overlap with each other. For non-trivial logic multiple inheritance is very fragile and error prone

### `poisson_sampling` flag was moved from `make_private` to `PrivacyEngineFactory`. 
The main principle I was following in this is that  each `PrivacyEngine` (including subclasses) defines a consistent behaviour, and `make_private` argument simply provide numerical parameters to configure this behaviour. I understand the downside of this: every parameter moved from `make_private` take some bit of flexibility away. From DP standpoint there's no reason we can't allow clients to use the same accountant or training with alternating clipping strategies. Ideally, we want to have more flexible interface for passing down accountant to enable multi-stage training (after all, accountant is the only state of the privacy engine object)

## ToDo
- tests for the factory


opacus/privacy_engine/factory.py
```diff
@@ -0,0 +1,33 @@
...
+    PrivacyEngine,
+    PrivacyEngineNonPoisson,
+    PrivacyEnginePerLayer,
+    PrivacyEnginePerLayerNonPoisson,
+)
+
+
+class PrivacyEngineFactory:
+    @classmethod
+    def get(
```
karthikprasad, line 12: nit: rename `get` to `create`?

opacus/privacy_engine/factory.py
```diff
@@ -0,0 +1,33 @@
...
+        cls,
+        accountant: str = "rdp",  # TODO: str or enum?
```
karthikprasad, line 14: I think Pytorch's style is to accept strings. (eg: loss_reduction); so I vote for str.
My very personal, subjective and highly controversial opinion: str with asserts are better than enums and their implicit checks.

opacus/privacy_engine/factory.py
```diff
@@ -0,0 +1,33 @@
...
+        clipping: str = "flat",  # TODO: str or enum?
+        poisson_sampling: bool = True,
+        secure_mode: bool = False,
+    ):
+        if accountant != "rdp":
+            raise NotImplementedError
```
karthikprasad, line 20: nit: add more detailed message

opacus/privacy_engine/catalog.py
```diff
@@ -0,0 +1,31 @@
+from opacus.privacy_engine.base import (
```
karthikprasad, line 1: nit: How about doing three imports: one to import dataloader, one to import privacyengine, and so on?

opacus/privacy_engine/factory.py
```diff
@@ -0,0 +1,33 @@
...
+class PrivacyEngineFactory:
+    @classmethod
+    def get(
+        cls,
+        accountant: str = "rdp",  # TODO: str or enum?
+        clipping: str = "flat",  # TODO: str or enum?
+        poisson_sampling: bool = True,
+        secure_mode: bool = False,
+    ):
+        if accountant != "rdp":
```
karthikprasad, line 19:
```suggestion
        if accountant.lower() != "rdp":
```

opacus/privacy_engine/factory.py
```diff
@@ -0,0 +1,33 @@
...
+            raise NotImplementedError
+
+        if clipping == "flat":
+            if poisson_sampling:
+                return PrivacyEngine(secure_mode)
+            else:
+                return PrivacyEngineNonPoisson(secure_mode)
```
karthikprasad, line 26: Very weak nit: is is better to have a single level switch instead of multi-level?
```suggestion
        if clipping == "flat" and poisson_sampling:
            return PrivacyEngine(secure_mode)
        elif clipping == "flat" and not poisson_sampling::
            return PrivacyEngineNonPoisson(secure_mode)
```

opacus/tests/batch_memory_manager_test.py
```diff
@@ -36,28 +38,28 @@ def _init_training(self, **data_loader_kwargs):
...
+        num_workers=st.integers(0, 4),
+        pin_memory=st.booleans(),
+    )
+    @settings(deadline=10000)
     def test_basic(
         self,
-        num_workers: int = 0,
-        pin_memory: bool = True,
+        num_workers,
+        pin_memory,
```
karthikprasad, line 49: why remove type hints?

karthikprasad (commented):
Awesome sauce @ffuuugor! Thanks for the very clear PR and the summary.
I added a few comments, mostly nits.

opacus/privacy_engine/factory.py
romovpa: `str` is the more pythonic way.
There is a [`Literal[]` type](https://docs.python.org/3/library/typing.html#typing.Literal) to check this. But it is introduced in py3.8. We can't afford use it right now, but we can add them later. Before than, asserts are a good solution. Or probably not just asserts, `if xxx: raise ValueError`.

romovpa: For the reference, [sklearn implementation](https://github.com/scikit-learn/scikit-learn/blob/793e90b344aeee90432b47515410c5666097f0e6/sklearn/linear_model/_logistic.py#L431) of `LogisticRegression(..., solver="lbfgs")`

opacus/privacy_engine/factory.py
romovpa: I would recommend not to do so. The potential value is that it is easier to make the code runnable. 
The downside is: This promotes inconsistency in the code. There is a change we will have a pain of handling these differences in different parts of the codebase in the future. It will be nearly impossible to roll back after people start using different spellings.

Also, try to find an error:
solver="Ibfgs"
(hint: correct is solver="lbfgs")

romovpa (commented):
@ffuuugor Thank you for visualizing the idea! 

I really like the way you implemented this via mixins. I left some comments on the appearance of the factory (specifically I suggest to make it less of a factory, more like a constructor).

Although I have some mixed feelings about this architecture. From one side, this implementation is easily extensible. From another side, this implementation of `make_private` is one step further from being self-explanatory on how DP is added using the low-level API (wrappers). Probably it shouldn't be self-explanatory (or maybe it is already, I just don't see it).

Can we wait couple more days before merging this? Possibly I will be able to put more constructive comments on this. 

This is the face of the library. This API will be called by most of the users. A lot is at stake here.

opacus/privacy_engine/base.py
```diff
@@ -57,6 +58,46 @@ def __init__(self, secure_mode=False):
...
             )
 
+    @abstractmethod
+    def _init_accountant(self) -> IAccountant:
+        pass
+
+    @abstractmethod
+    def _prepare_optimizer(
+        self,
+        optimizer: optim.Optimizer,
```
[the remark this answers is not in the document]
ffuuugor: Hmm,
I remember having a todo about this for `make_private` method - it's still there.

opacus/privacy_engine/base.py
```diff
@@ -242,20 +221,71 @@ def make_private_with_epsilon(
...
+        module = super()._prepare_model(module, batch_first, loss_reduction)
+        module.register_forward_pre_hook(forbid_accumulation_hook)
 
-        # wrap
-        if isinstance(module, GradSampleModule):
-            return module
+        return module
+
+
+class SequentialBatchDataLoaderMixin:
```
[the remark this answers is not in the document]
ffuuugor: Sure, docstrings are a must, especially for stuff like this.
I'll make sure to have them, but probably in a separate PR once we've settled with the architecture

opacus/privacy_engine/factory.py
ffuuugor: >If I understand correctly the idea, the class is redundant here. It's just a function.
Also Factory.get to my taste looks a little bit like a Java developer invaded the building.

Guilty as charged. You're right, this looks like a hardcore OOP pattern, we don't really need in here

> This get() looks like a constructor part absent in the implementation classes. If so, why not doing something like:

I'm not sure I understand what you mean by that.

Regardless, the subclass router option you've proposed look good to me. I'd be fine with either `build_privacy_engine()` function (replacing current factory) or subclass router in PrivacyEngine constructor.
I'm not sure I see a strong arguments for either - both looks ok?

cc @karthikprasad @alexandresablayrolles

opacus/privacy_engine/factory.py
ffuuugor: Thanks!

opacus/privacy_engine/catalog.py
ffuuugor: for visible logical grouping?

opacus/tests/batch_memory_manager_test.py
ffuuugor: went a little too far when removing default values, thanks for noticing

opacus/privacy_engine/factory.py
karthikprasad: +1 to @romovpa's call out. I vote for a function to replace the factory and its method. A nit on the naming: `get_privacy_engine()` instead of `build_privacy_engine()` :)

> I'm not sure I see a strong arguments for either - both looks ok?

If not going ahead with factory pattern, I have a stronger argument in favour of function. The class here is neither acting on instance specific data with associated methods, nor is it a namespace; class doesn't add any value.

opacus/privacy_engine/catalog.py
karthikprasad: Yup. Just for readability.

opacus/privacy_engine/factory.py
karthikprasad: Touché. I withdraw my comment.

opacus/privacy_engine/factory.py
karthikprasad: TIL about Literal[] type. Die enum die!

ffuuugor:
>Can we wait couple more days before merging this? Possibly I will be able to put more constructive comments on this.
>This is the face of the library. This API will be called by most of the users. A lot is at stake here.

Yes, absolutely, 100% agree here. I myself am not perfectly satisfied with how it looks, and totally agree this is a very important bit of the API.

The goals I wanted to achieve
- Support per layer clipping strategy. While not super beneficial by itself for model convergence or privacy, it helps a lot for efficient parallelization (see #191)
- Support for different accountants. In addition to what we already have implemented, there are few ideas interesting to play with: https://arxiv.org/pdf/2106.02848.pdf, https://arxiv.org/pdf/2008.11193.pdf or DP-FTRL
- Do it in a way that's extensible for future use cases

> From another side, this implementation of make_private is one step further from being self-explanatory on how DP is added using the low-level API (wrappers). Probably it shouldn't be self-explanatory (or maybe it is already, I just don't see it).

I agree that readability is an important goal here.
I can argue, PrivacyEngineBase:make_private is still clean and clear, and also have everything you need to know: the only changes coming from elsewhere are accumulation forbidding hook (poisson mixin) and distributed per layer hooks.
For me the most important aspect is readability for the default use case, and it's mostly maintained.

However, it's true that readability overall took a hit with this mixin implementation: you now need to look at multiple places to understand what's going on. Even worse, you need to understand how those mixins work together - the reason people generally disfavour multiple inheritance.

karthikprasad:
> However, it's true that readability overall took a hit with this mixin implementation: you now need to look at multiple places to understand what's going on.

Resurfacing the wild idea! How about just have multiple PrivacyEngine subclasses for every combination. Leads to code duplication, yes. But a lot more readable; and also sends a clear recommendation to the user that custom implementations need to be achieved via subclassing (but most of the code will be copy-pasta so it won't be hard)

opacus/privacy_engine/factory.py
ffuuugor: Yup, that's a good argument - I'll switch to function.

ffuuugor:
> How about just have multiple PrivacyEngine subclasses for every combination

I mean, we do subclasses, we just use mixins to avoid copy-pasta.
I'd vote against relying on a copy-pasta as a foundational design principle. While more readable, it'd be a huge pain to maintain - even now we've had troubles keeping `make_private` and `make_private_per_layer` in sync. With more options it'd be exponentially worse

romovpa:
> Support for different accountants

Noob question. Why don't we run all available accountants in parallel? They look cheap, we could have all possible DP estimates at once. No choice, no problem. 

And another (probably stupid) question. Why do we need to set alphas in RDP? If I understand correctly, by changing alpha we get different upper bounds of eps. Is it possible to find the optimal combination every time?

cc @alexandresablayrolles

romovpa:
> Do it in a way that's extensible for future use cases

Do we have the list of real and imaginary, future and present use cases? This may aid our design decisions. 
If not, I would recommend to create one and refresh it time to time.

ffuuugor:
>Noob question. Why don't we run all available accountants in parallel? They look cheap, we could have all possible DP estimates at once. No choice, no problem. 

We can run multiple, but the list is never exhaustive. Ideally, people should be able to plug their own.

Additionally, there is once consideration, which I now regret - it might have brought more troubles than good. There's an idea of data-dependant per-user privacy accounting https://arxiv.org/abs/2008.11193, which would call for a different accountant interface, not only implementation (current accountants take (noise_multiplier, sample_rate) on each step, data-dependant will also require grad norms)

> Why do we need to set alphas in RDP?
Not a stupid question at all.
I can walk you through Renyi DP offline, but a very short version is:

1) There's a notion of Renyi Divergence with alpha as a parameter
2) Renyi Divergence is then used for Renyi DP definition: for every alpha you can certain epsilon, which together give you (alpha, eps)-Renyi DP
3) RDP then can be translated into regular (eps, delta)-DP using a simple transformation. The interesting part is that eps-from-normal-DP will depend on eps from RDP, target delta AND alpha from RDP. 

So yes, it is, after all, about finding the best value of alpha to get the tightest bound on (eps, delta)-DP, but as far as I'm aware there's no way to analytically solve this - that's why we do brute-force search over an array of alphas.

ffuuugor:
In an ironic turn of events, I just realised, that we don't need PrivacyEngine inheritance for the use cases we have so far

a) Data dependant accountant would've required different API and therefore inheritance, but we have no firm plans to support it (and definitely not before the launch)
b) per layer clipping + distributed did require this, but can be easily refactored to avoid it

I really hate last-minute changes to the core API, but it looks like we can either

a) avoid inheritance altogether, pass accountant, clip_strategy and is_poisson as parameters
b) keep very minimalistic inheritance, where the only thing subclasses change are DPOtimizerClass and _get_data_loader() method + "get_privacy_engine" method as @karthikprasad  proposed

I am inclined towards the second, as it leaves more freedom for us in the future.

a) Keep different clippings in different PrivacyEngineSubclasses
b) take accountant as a PrivacyEngine constructor parameter. It has two side bonuses: easy to switch for another + people can pass non-empty accountants to continue tracking privacy in multi-stage training.
c) not sure about Poisson: does look better as a mixin, contructor parameter or make_private parameter?

romovpa:
> I really hate last-minute changes to the core API

@ffuuugor But it's the most important change :)

Let's put all the craziest ideas here. If we don't find the best solution, at least we will have them documented so we can consider them later.

ffuuugor:
Ok, here goes another prototype

1) accountant is passed to the PrivacyEngine constructor. It was proposed before, but I was opposed to the idea, because it doesn't scale well to some type of accountants, which require different call interfaces (e.g. data-dependant accountant).

Since then I realised, that this is probably a premature concern, and we can leave future us to deal with it. We can always subclass PrivacyEngine for new type of accountant if we decide to have one. Both current accountants fit into the single interface.

2) _prepare_optimizer is consolidated between PerLayer and normal clipping.
The only difference (apart from the optimizer class) was `optimizer.register_hooks(module)` for DistributedDPPerLayerOptimizer. This is not strictly necessary: we don't need to pass `module` to the call, and therefor can do the same in optimizer's constructor

karthikprasad:
All points well taken @ffuuugor. I am all in favour of passing args instead of inheritance. My concerns were and are the following:
<begin rant/>
1. The args may not be compatible for every combination. Eg; will there be `Accountant1` which won't work with `DataLoader2` which won't work with `Clipper3`? If yes, adding these compatibility checks in the PrivacyEngine is both ugly and unscalable.
2. Personally, I prefer, having just one way to do custom implementation. Or at the very least, provide a strong recommendation on the right way to do things even if we support multiple approaches.
    * With args approach, I fear we might end up saying some cases are easier by passing args and for some subclassing is better for some others. Except, we are not (and can't) explicitly saying what those "some" cases are.
    * This is also the reason why I suggested only supporting sub-classing rather than subclassing AND mixins+factory. However, this is still not bad since we can "recommend" folks to do subclassing and factory is reserved for Opacus approved combos only.
        * Obviously, we use python so people can do what they want, but not heeding to recommendation might land them in bad soup.

</end rant>

Looking at your comments, it seems like your proposal won't run into the above issues, but I'm unable to see how exactly.

ffuuugor:
@karthikprasad  thanks a lot for your feedback, that's super valuable.
tbh, I'm more in a state "we have to choose between two bad options"

> The args may not be compatible for every combination

That would indeed be a job better suited for some privacy method. Do you expect something like this? It's not the case for the current set of strategies and I'm somewhat hesitant to build a design based on possibilities (it tends to be unnecessarily complex).

> Personally, I prefer, having just one way to do custom implementation

Let's split this into two conversations: users are supposed to do, and what we're going to do in the future.

1. For built-in opacus features I see a future where we adopt factory approach at some point, when we clearly see the need for it. Now we didn't, because all accountants and optimizers have exactly the same interface, and don't require complex coordination logic. This might very well not be the case in the future. But there's nothing wrong with this - we find a new feature, we introduce new design to support it, while graciously deprecating the old one
2. For custom code by users inheritance is still the way to go. 

> Looking at your comments, it seems like your proposal won't run into the above issues, but I'm unable to see how exactly.

You can now also check out the code, it's there

karthikprasad (commented):
> You can now also check out the code, it's there

Ouch! I missed the fact that you had already pushed the commits. Apologies. My rant was futile. :P

> Do you expect something like this? It's not the case for the current set of strategies and I'm somewhat hesitant to build a design based on possibilities (it tends to be unnecessarily complex).

I don't expect it anytime soon, but then again, I don't know if I should trust my expectations. My concern was: should there be one such accountant, the current API won't be robust.

Question: What is the motivation behind making accountant an arg to constructor? Do we have a lot of accountants that can work out of the box with the current engine without us having to check/validate anything?
If yes, this design looks good! If no, then why not treat accountant similar to optimizer? 

What do I mean by "similar to optimizer"?
Optimizer is not passed as an arg but is instead constructed based on str args (flat and distributed). If a new approach (say per-tensor) arrives, this would necessitate sub-classing by user OR an additional switch in the if-else block in the main privacy engine, right? 

Similarly, I propose we have `_prepare_accountant` which accepts a str and does something similar
```
if "rdp" == accountant:
  return RDPAccountant()
else "moments" == accountant:
  return MomentsAccountant()
```
If the user wants to use a custom accountant, they can subclass. If we want to natively support it, we can add it as another ladder in the if-else construct. WDYT?

ffuuugor:
>  My concern was: should there be one such accountant, the current API won't be robust.

Is the  concern here that users will pass something else as an accountant and unknowingly get incorrect accounting?


> What is the motivation behind making accountant an arg to constructor?

The initial motivation why I liked this approach was that it would've allowed for something like this:
```
accountant = RDPAccountant()
privacy_engine = PrivacyEngine(accountant=accountant)
<...train...>
privacy_engine = SomeNewPrivacyEngine(accountant=accountant)
<...continue training...>
```

That would've given a great flexibility to maintaining privacy budget across multiple steps of training.
However, now that I've laid it out, I see a big gap in this approach;
First, in the current implementation it doesn't make sense to create new PrivacyEngines: every important parameter is passed to a `make_private` call. So there's no reason to create another PrivacyEngine (just to be clear, that isn't the case if we're doing subclassing)
Second (and probably more important): if we're passing around accountant, we're completely abdicating the responsibility of keeping privacy budget *per dataset*. We have a TODO to handle consecutive calls to `make_private` with data_loaders over different datasets - with the use case I've described above we won't be able to do anything about it.

In conclusion, I agree with your proposal for doing this "similar to optimizer" and will implement it.

> Do we have a lot of accountants that can work out of the box with the current engine without us having to check/validate anything?

In reality, I think that people writing their own accountants will find a way to inject it (there will be few, and there is relatively easy way to do it), and vast majority should use accountants provided by the core library.

Thanks @karthikprasad, that was very useful feedback

opacus/optimizers/ddp_perlayeroptimizer.py
```diff
@@ -92,12 +92,14 @@ def _ddp_per_layer_hook(
...
 
-    def register_hooks(self, module: GradSampleModule):
-        module.ddp_hooks = []
+    def register_hooks(self):
         for p, max_grad_norm in zip(self.params, self.max_grad_norms):
             if not p.requires_grad:
                 continue
 
-            module.ddp_hooks.append(
+            if not hasattr(p, "ddp_hooks"):
```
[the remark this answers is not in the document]
ffuuugor: ha, they actually are related to PrivacyEngine.

Before these changes `DistributedPerLayerOptimizer` required special handling in `make_private` method: 

```
optimizer.register_hook(model)
```

This call was unique for DistributedPerLayerOptimizer because of the optimization that only kicks in when both distributed training and per layer clipping is enabled. (see #196)

With these changes we can ditch an extra call (move it to optimizer's `__init__`) and provide unified `make_private` implementation for all optimizers

opacus/privacy_engine.py
```diff
@@ -80,6 +83,85 @@ def __init__(self, secure_mode=False):
...
+            optimizer = optimizer.optimizer
+
+        generator = None
+        if self.secure_mode:
+            generator = self.secure_rng
+        elif noise_seed is not None:
+            generator = torch.Generator()
+            generator.manual_seed(noise_seed)
+
+        optim_class = self._get_optimizer_class(
```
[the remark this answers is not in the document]
ffuuugor: why?

karthikprasad:
> we're completely abdicating the responsibility of keeping privacy budget per dataset. 

Ah! Good catch! I had not considered this and hence was not very favourable to my own recommendation. Now, with this point, it is easier for me to get behind it.

> In conclusion, I agree with your proposal for doing this "similar to optimizer" and will implement it.

You might want to hold on implementing just yet. @romovpa has requested for a couple of more days for constructive comments on the design and he might have a better suggestion.

opacus/privacy_engine.py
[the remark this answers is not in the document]
karthikprasad: > why?

For semantic clarity.

opacus/accountants/gdp.py
```diff
@@ -3,25 +3,31 @@
...
         if noise_multiplier != self.noise_multiplier or sample_rate != self.sample_rate:
             raise ValueError(
                 "Noise multiplier and sample rate have to stay constant in GaussianAccountant."
             )
         self.steps += 1
 
-    def get_privacy_spent(self, delta: float) -> float:
-        if self.poisson:
+    def get_epsilon(self, delta: float, poisson: bool = True) -> float:
+        if poisson:
```
alexandresablayrolles, line 25: Why move poisson here?

opacus/privacy_engine.py
```diff
@@ -80,6 +79,92 @@ def __init__(self, secure_mode=False):
                 "one last time before production with ``secure_mode`` turned on."
             )
 
+    def _prepare_accountant(self, mechanism: str) -> IAccountant:
```
alexandresablayrolles, line 82: Should we move that to accountants/__init__.py? In any case it is a "static" method.

opacus/privacy_engine.py
```diff
@@ -80,6 +79,92 @@ def __init__(self, secure_mode=False):
...
+        if mechanism == "rdp":
+            return RDPAccountant()
+        elif mechanism == "gdp":
+            return GaussianAccountant()
+
+        raise ValueError(f"Unexpected accounting mechanism: {mechanism}")
+
+    def _get_optimizer_class(self, clipping: str, distributed: bool):
```
alexandresablayrolles, line 90: Same idea: move it to optimizers/__init__

opacus/privacy_engine.py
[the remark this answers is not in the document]
alexandresablayrolles: +2 (+1 for semantic clarity and +1 to actually move this function where it belongs, see my comment above)

alexandresablayrolles (changes requested):
Looks great to me overall! My main point is whether we should move _get_*_class to optimizers/ and accountants/

opacus/accountants/gdp.py
alexandresablayrolles: Got it actually! Seems necessary to not have any extra parameters in the constructor

opacus/privacy_engine.py
ffuuugor: Hmm, that is an excellent point

opacus/privacy_engine.py
```diff
@@ -57,7 +52,7 @@ def __init__(self, accountant: str = "rdp", secure_mode=False):
             ...         optimizer.virtual_step()  # this will call privacy engine's virtual_step()
 
         """
-        self.accountant = self._prepare_accountant(mechanism=accountant)
+        self.accountant = get_accountant(mechanism=accountant)
```
karthikprasad, line 55: I thought the plan was to move this to `make_private` similar to optimizer?

opacus/privacy_engine.py
```diff
@@ -276,6 +249,7 @@ def make_private_with_epsilon(
             module=module,
             optimizer=optimizer,
             data_loader=data_loader,
+            # TODO: what if the client has selected non-RDP accountant?
```
karthikprasad, line 252: This can be handled if accountant is not fixed during construction and instead passed here as arg.

opacus/privacy_engine.py
ffuuugor: I would disagree - keeping accountant across multiple calls to `make_optimizer` is valuable and allows you to calculate overall epsilon over multi-stage training (e.g. if you want to change batch_size in the middle of your training)

opacus/privacy_engine.py
ffuuugor: The problem here is that we don't yet have reverse epsilon computation for Gaussian DP (i.e. method to get noise multiplier given the target epsilon)

opacus/privacy_engine.py
karthikprasad: Good point!

alexandresablayrolles (approved):
Great diff thanks!

karthikprasad (approved):
giphy

PR #255, opened 2021-11-10
Fix lint

alexandresablayrolles:
Fix lint


ffuuugor (approved):
Thank you! We're one merge away from having all tests green on 1.0 for the first time!

PR #257, opened 2021-11-10
Fixes

alexandresablayrolles:
Removing drop_last as it is incompatible with a sampler + updating DDPOptimizer for batchsplitter.

### DIFF
--- a/opacus/optimizers/ddpoptimizer.py
+++ b/opacus/optimizers/ddpoptimizer.py
@@ -7,7 +7,6 @@ from torch.optim import Optimizer
 
 from .optimizer import DPOptimizer
 
-
 class DistributedDPOptimizer(DPOptimizer):
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


ffuuugor (approved):

opacus/optimizers/ddpoptimizer.py
```diff
@@ -47,7 +46,8 @@ def reduce_gradients(self):
                 p.grad /= self.world_size
 
     def step(self, closure: Optional[Callable[[], float]] = None) -> Optional[float]:
```
romovpa, line 48: Are we sure the output should be `float` and not `Tensor`? 

In this [example](https://pytorch.org/docs/stable/optim.html):
```python
for input, target in dataset:
    def closure():
        optimizer.zero_grad()
        output = model(input)
        loss = loss_fn(output, target)
        loss.backward()
        return loss
    optimizer.step(closure)
```

Here `loss` is obviously a `Tensor`, we can access grad and call `backward()`.

alexandresablayrolles: Fixed, thanks!

PR #258, opened 2021-11-11
[Proposal] Remove DP prefix from opacus.layers.* as they don't carry any DP logic

romovpa:
_I found this interesting and I want to share this with you. But the impact is low, I'm not going to push for it. If it gets support, it is better to make this move before v1.0 release._

### What is opacus.layers for
The layers from `opacus.layers` are intended to be **drop-in compatible with their `torch.nn` analogs**. 
- They intentionally do not carry any logic related to any part of the DP-SGD. 
- They are designed to be seamlessly interchangeable with the `nn` analogs outside of the Opacus world.
- They are practically a workaround for the problem that we cannot efficiently introspect and hook the native implementation.

### Let's remove DP prefix

If there is no link to DP, why not:
- Remove DP prefix from the class names and give them their original names: `dp.LSTM` instead of `DPLSTM`
- Use a shortcut to access DP-friendly modules: `import opacus.layers as dp` by analogy with `from torch.nn import nn`
- Recommend all public classes in `opacus.layers` to have the same **name**/interface/state_dict as the corresponding modules from `torch.nn`.  
- Rename opacus.layers.* subpackages to match `torch.nn.*` names.

Example:
```python
import torch.nn as nn
import opacus.layers as dp

model_nn = nn.LSTM(input_size, hidden_size)  # Original
model_dp = dp.LSTM(input_size, hidden_size)  # DP-friendly
```

Alternative shortcut:
```python
import opacus.layers as dp_layers
import opacus.layers as dp_nn
import opacus.dp_layers as dp
```

### Pros

- No false impression that DP* layers have a specific logic related to DP.
- Emphasize that the modules from `opacus.layers` are identical to those from `nn`. 
- Less clutter. (`DPLSTM` vs `dp.LSTM`)
- Easier deprecation of `dp.*` classes in favor of the original implementations when we grad sampler is smart enough to deal with them.
- The package can be reused outside of Opacus

### Cons

- Easy to get confused if the modules are accessed directly, not via `{nn/dp}.Module`. 
    - The risk can be reduced performance when unintentionally using dp module instead of nn one.
- Not clear from the repr-string which exactly module is used. Prefix in the name communicates to the user that they are essentially using a very different module when the print out the model with print (model)


ffuuugor:
You get my ~sword~ vote!

karthikprasad:
Um! Sorry @romovpa but I am tempted to snatch the sword given to you by @ffuuugor back. Please feel free to tie me down by poking holes in my thoughts.

The main reason for the different name is communicate to the user that they are essentially using a very different module when the print out the model with `print(model)`. Even though the modules do the same job and have the same interface and state_dicts, their implementations are different. Compare this to other modules such as `Linear`, where even the implementations are the same; the only change there being addition of hooks to capture `.grad_samples`

As for the prefix `DP`, I agree with you it is not the best prefix. Perhaps change it to `GSLSTM` or `OpacusLSTM`?

alexandresablayrolles:
+1 for @karthikprasad's proposal of having a "GS" prefix.

romovpa:
Thanks @karthikprasad! The need to indicate the difference in the repr-string of the class is a fair argument. I've added this to Cons in the description.

Between "DP" prefix and "GS" I personally prefer "DP". Just to reduce the perplexity. I believe for not very experienced users it would be easier to remember and understand "DP" despite the fact that it can create false impression; "GS" is a new thing to learn.

Let's leave it for now then.

PR #260, opened 2021-11-13
Fix for floating point representation attack

ashkan-software:
## Types of changes

- [x ] Fix for a possible attack

Gaussian mechanism for differential privacy is susceptible to a similar floating point attack against Gaussian noise sampling as the one for Laplace noise as proposed by Mironov in CCS 2011 (https://www.microsoft.com/en-us/research/wp-content/uploads/2012/10/lsbs.pdf)). 

The attack is possible since not all real numbers can be represented in floating points, hence not all can be sampled either.  The attack is possible since normal noise distribution implementation of pytorch is used.  If one observes an output that is protected by Gaussian mechanism, they can determine the noiseless value, hence, invalidating differential private guarantees.

This PR fixes this issue by calling the Gaussian noise function 2*n times, when n=2 (see section 5.1 in https://arxiv.org/abs/2107.10138)

I ran 3 benchmarks to see if this change impacts the running time, as we call the Gaussian noise 4 times. It has a mild impact on the running time 😁 I ran IMDB on my laptop. I ran MNIST on Google collab. I ran vision benchmark on Google collab with GPU. Here are the summary of those runs (I average over 5 runs each):

Before the change:
MNIST: 2m17s
IMDB: 3m0s
Vision benchmark: 2m9s

Before the change:
MNIST: 2m27s
IMDB: 3m16s
Vision benchmark: 1m48s !!!

commands used to run these:
```
!time python3 examples/mnist.py --device=cpu --n=3 --lr=.25 --sigma=1.3 -c=1.5
!time python3 examples/imdb.py --device=cpu --epochs=3 --lr=.25 --sigma=1.3 -c=1.5
!time python3 examples/vision_benchmark.py --device=cuda --lr=.25 --sigma=1.3 -c=1.5 --batch-size=64
```

## Checklist

- [ ] All tests passed, and additional code has been covered with new tests.


opacus/optimizers/optimizer.py
```diff
@@ -8,19 +8,60 @@
...
+            for generating the noise
+        generator: The PyTorch noise generator
+        secure_mode: boolean showing if "secure" noise need to be generate
+            (see the notes)
+
+    Notes:
+        If `secure_mode` is enabled, the generated noise is also secure
+        against the floating point representation attacks, such as the ones
+        in https://arxiv.org/abs/2107.10138. This is achieved through calling
+        the Gaussian noise function 2*n times, when n=2 (see section 5.1 in
```
karthikprasad, line 31: curious, why is `n` set to 2?

opacus/optimizers/optimizer.py
```diff
@@ -8,19 +8,60 @@
...
+    zeros = torch.zeros(reference.shape, device=reference.device)
+    if std == 0:
+        return zeros
+    # TODO: handle device transfers: generator and reference tensor
+    # could be on different devices
+    if secure_mode:
+        torch.normal(
+            mean=0,
+            std=std,
+            size=(1, 1),
```
karthikprasad, line 43: Dumb question: is it OK for the shape to be something different? I'm guessing yes, since you explicitly coded it this way, but wanted to confirm.

ffuuugor: This call to `torch.normal` doesn't seem to have any effect - the result it not stored or used anywhere, unless I've missed something

opacus/optimizers/optimizer.py
```diff
@@ -8,19 +8,60 @@
...
+            device=reference.device,
+            generator=generator,
+        )  # throw away first generated random number
+        sum = zeros
+        for i in range(4):
```
ffuuugor, line 48: I trust you and Ilya that it solves the problem, but could you pls do ELI5 why this works? 
I remember "Option 3" from your doc, but I'm not sure I understand how this relates to it

opacus/tests/randomness_test.py
```diff
@@ -178,6 +178,7 @@ def _init_training(self, generator, noise: float = 1.0):
             max_grad_norm=1.0,
             expected_batch_size=8,
             generator=generator,
+            secure_mode=False,
```
ffuuugor, line 181: nit: You can just leave it as it is, since `False` is default value

ffuuugor (changes requested):
Thanks for this!
Question on the benchmarking - do you have any intuition around vision_benchmark results? How many runs did you do - could it just be random fluctuations?

As a general comment, we now have a test for the noise lever (privacy_engine_test.py:test_noise_level), can you pls update it to cover secure_mode?

opacus/optimizers/optimizer.py
alexandresablayrolles: The trick that the distribution of the squared norm of a Gaussian is exponential. (see second sentence in "Gamma, exponential and related" in https://en.wikipedia.org/wiki/Chi-squared_distribution).
Thus, you can sample from a 2D gaussian by sampling first the norm (log(U) if U is uniform gives you exponential) and then the angle (uniform between 0 and 2Pi). That thus gives you two independent samples from the gaussian.

opacus/optimizers/optimizer.py
alexandresablayrolles: From what I understand, the sum from 1 to 4 gives you a Gaussian with variance 4 std^2, thus sum/2 is a gaussian with variance std^2. @ashkan-software : shouldn't you loop over only 2 samples as per the docstring?

opacus/optimizers/optimizer.py
ashkan-software: So basically `n` can be any number > 1. The bigger, the more computation needs to be done (`2n` Gaussian samples will be generated). The reason we chose `n=2` is that, `n=1` could be easy to break and `n>2` is not really necessary. Let me explain: the complexity of the attack is `2^p(2n-1)`. In PyTorch, `p=53` and so complexity is `2^53(2n-1)`. With `n=1`, we get `2^53` (easy to break) but with `n=2`, we get `2^159`

opacus/optimizers/optimizer.py
ashkan-software: Sorry for the confusion.

The purpose of this line of code is to generate one Gaussian sample and throw away (that's why it's not used).

According to the paper: "Readers are encouraged to *discard* the first result from Box-Muller before sampling in this way, to ensure no carry-over of uniform variates from a previous draw, which may have already been successfully attacked."

ashkan-software: I will make the comment more clear

opacus/optimizers/optimizer.py
```diff
@@ -8,19 +8,60 @@
...
+        against the floating point representation attacks, such as the ones
+        in https://arxiv.org/abs/2107.10138. This is achieved through calling
+        the Gaussian noise function 2*n times, when n=2 (see section 5.1 in
+        https://arxiv.org/abs/2107.10138)
+    """
+    zeros = torch.zeros(reference.shape, device=reference.device)
+    if std == 0:
+        return zeros
+    # TODO: handle device transfers: generator and reference tensor
+    # could be on different devices
```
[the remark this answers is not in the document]
ashkan-software: I just leave it as is for now :)

opacus/optimizers/optimizer.py
ashkan-software: Great question. 

This approach is actually not any of those 3 options we had. I found this approach in a recent paper and me and Ilya think this is an intelligent way of fixing the problem. The idea is to invert the Gaussian mechanism and guess what values used as input to the mechanism. This is possible if Gaussian method is used once. But if we use the Gaussian more than once (in this fix, we call it 4 times), it becomes exponentially harder to guess those values. This is in very simple words, but the fix is a bit more involved and is explained in the paper I listed on the PR

opacus/optimizers/optimizer.py
ashkan-software: @alexandresablayrolles not sure if I understand what you explained, but I hope @karthikprasad got his response? :)

ashkan-software:
@ffuuugor the only intuition I have for the vision dataset is that I have made a mistake in reporting :) haha. The correct reporting should be:

Before the change:
Vision benchmark: 1m47s

Before the change:
Vision benchmark: 1m48s

opacus/optimizers/optimizer.py
ashkan-software: @alexandresablayrolles 

The reason for having number 4 and 2 in the code, is that when `n=2`, we get those values (Section 5.1 in the paper):

`sum(gauss(0, 1) for i in range(2 * n)) / sqrt(2 * n)`

opacus/optimizers/optimizer.py
ilyamironov: > The trick that the distribution of the squared norm of a Gaussian is exponential. (see second sentence in "Gamma, exponential and related" in https://en.wikipedia.org/wiki/Chi-squared_distribution). Thus, you can sample from a 2D gaussian by sampling first the norm (log(U) if U is uniform gives you exponential) and then the angle (uniform between 0 and 2Pi). That thus gives you two independent samples from the gaussian.

@alexandresablayrolles You are right ... if the computations were done over the reals. Unfortunately, the two samples are *not* independent if computed with floating-point arithmetic. What we do instead is using the fact that a sum of Gaussians is a Gaussian, and transform two (or four) subtly correlated Gaussians into one with much smoother distribution.

opacus/optimizers/optimizer.py
karthikprasad: The fact that you were throwing it away was clear to me. What was not clear is if the shape can be any dumb thing. I am not familiar with how `torch.normal` is implemented and if using a different shape/size than the subsequent has any impact.

ashkan-software: Oh I see. The shape argument in `torch.normal` is to determine how many random numbers to generate. I am here generating one number by doing `(1,1)` @karthikprasad

ashkan-software:
I am now working on the `privacy_engine_test.py:test_noise_level` that Igor suggested. This test is failing here, but passes locally on my computer. Any ideas?

This is the output of the test here in this PR

```

_________________ PrivacyEngineConvNetTest.test_noise_level ___________________

self = <opacus.tests.privacy_engine_test.PrivacyEngineConvNetTest testMethod=test_noise_level>

>   ???

opacus/tests/privacy_engine_test.py:338: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <hypothesis.core.StateForActualGivenExecution object at 0x7f9b97abcf10>
message = 'Hypothesis test_noise_level(self=<opacus.tests.privacy_engine_test.PrivacyEngineConvNetTest testMethod=test_noise_lev...804311150971176, max_steps=8) produces unreliable results: Falsified on the first call but did not on a subsequent one'

    def __flaky(self, message):
        if len(self.falsifying_examples) <= 1:
>           raise Flaky(message)
E           hypothesis.errors.Flaky: Hypothesis test_noise_level(self=<opacus.tests.privacy_engine_test.PrivacyEngineConvNetTest testMethod=test_noise_level>, noise_multiplier=0.6804311150971176, max_steps=8) produces unreliable results: Falsified on the first call but did not on a subsequent one

../.local/lib/python3.8/site-packages/hypothesis/core.py:893: Flaky
```

ashkan-software:
I debugged this issue further, and it seems like this test is flaky: it passes some times and fails some other times. I am simply rerunning it on my computer - it sometimes fails, and sometimes pass. To me it also makes sense, because it is written to validate the output of a random number generator. So it depends on what value is.

ffuuugor:
>I debugged this issue further, and it seems like this test is flaky: it passes some times and fails some other times

Note, that we fix the seed in the beginning of the test - so the noise generated should be consistent across machines.

Also note, that sometimes (especially on CircleCI) tests fail due to timeout (you should set `deadline=None` instead of `deadline=20000`)

0.05 -> 0.1 change was to make it less flaky. The main point of this test is not to check we're doing it correctly now, but to keep checking it in the future so that we don't accidentally break it. The intuition is that we know, that we're adding the correct amount of noise now -> flaky test results indicate problems with the test, not the code

ashkan-software:
Update: I added a new test to ensure the sum of 4 gaussians and division by 2 is not changing in the future. I also changed `test_noise_level` to check for both when `secure_mode` is off/on. Github for some reason shows the changes in a very complicated way, but the changes are minimal :)

ashkan-software:
@ffuuugor 
> 0.05 -> 0.1 change was to make it less flaky. The main point of this test is not to check we're doing it correctly now, but to keep checking it in the future so that we don't accidentally break it. The intuition is that we know, that we're adding the correct amount of noise now -> flaky test results indicate problems with the test, not the code

I now changed 0.1 > 0.15 😂 so that the flaky test passes. It failed sometimes and sometimes it passed 😂
Not really sure what this test is checking, if you say this is not checking we are doing something correctly lol. You wrote 
>but to keep checking it in the future so that we don't accidentally break it. 

but seems like we are not really checking strictly, so in the future we may actually break this!

opacus/optimizers/optimizer.py
```diff
@@ -8,19 +8,68 @@
...
+        )  # generate, but throw away first generated Gaussian sample
+        sum = zeros
+        for i in range(4):
+            sum += torch.normal(
+                mean=0,
+                std=std,
+                size=reference.shape,
+                device=reference.device,
+                generator=generator,
+            )
```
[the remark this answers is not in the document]
ffuuugor: From the code readability perspective I'd rather prefer the loop.
I can see how your proposal can be beneficial for the performance, but if it's not, I think loop makes it more clear

opacus/tests/privacy_engine_test.py
```diff
@@ -338,47 +340,84 @@ def test_deterministic_run(self):
...
-            / (optimizer.expected_batch_size ** 2)
+        helper_test_noise_level(
+            noise_multiplier=noise_multiplier,
+            max_steps=max_steps,
+            secure_mode=False,
+        )
+        helper_test_noise_level(
+            noise_multiplier=noise_multiplier,
+            max_steps=max_steps,
+            secure_mode=True,
```
ffuuugor, line 399: nit: nice syntax that can help future debugging:

```
with self.subTest(secure_mode=True):
    helper_test_noise_level(...)

with self.subTest(secure_mode=False):
    helper_test_noise_level(...)
```

Doesn't do much, but provides nice summary what exactly failed

ffuuugor (approved):
Thank you, awesome stuff!

opacus/tests/privacy_engine_test.py
ashkan-software: Cool! I will include this!

opacus/optimizers/optimizer.py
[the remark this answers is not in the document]
ashkan-software: So I changed this to what @ilyamironov suggested and ran the tests. The runtime is the same (even slightly more). So I leave the for-loop.

PR #261, opened 2021-11-15
Enable validation for GradSampleModule

ffuuugor:
## Substantial changes

1) GradSampleModule now checks the input module for compatible modules.
Motivation: if using `GradSampleModule` outside `PrivacyEngine`, we want to signal to the user is their module contains unsupported or incompatible modules. The impact of not doing so would be either unexpectedly missing `grad_samples` on some parameters (remember, the user might not know the exact list of supported modules) or wrong results altogether (if there's something like `BatchNorm` is involved)

2) We switch to simple model for `grad_sample_module_test.py`. The reason is `mobilenet_v3_small` actually contains `BatchNorm` and don't pass validation

### DIFF
--- a/opacus/grad_sample/grad_sample_module.py
+++ b/opacus/grad_sample/grad_sample_module.py
@@ -2,6 +2,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
 from __future__ import annotations
 
+import logging
 from functools import partial
 from typing import List, Tuple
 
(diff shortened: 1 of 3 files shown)


opacus/grad_sample/grad_sample_module.py
```diff
@@ -59,8 +63,24 @@ class GradSampleModule(nn.Module):
...
     GRAD_SAMPLERS = {}
 
-    def __init__(self, m: nn.Module, *, batch_first=True, loss_reduction="mean"):
+    def __init__(
+        self,
+        m: nn.Module,
+        *,
+        batch_first=True,
+        loss_reduction="mean",
+        strict: bool = True,
```
karthikprasad, line 72: I like this arg name. I'll use this in ModuleValidator too instead of `raise_if_error`! :)

karthikprasad (approved):

PR #264, opened 2021-11-17
address TODOs, part 1

ffuuugor:
Addressing various ToDo comments across the codebase.

### mandatory `zero_grad`

Not calling zero_grad in most scenarios breaks privacy accounting. 
We want to prohibit this behaviour (after all, that's probably the most common mistake people do in their training pipelines: forgetting zero_grad).
However, ideally we don't want to do anything surprising.

Options considered and discarded:
1) Call `self.zero_grad()` inside `optimizer.step()`.
  a) Doesn't work with virtual steps
  b) Unexpected optimizer behaviour - normal optimizers don't do that
  c) Limited debugging access: users loses access to gradient tensors (they no longer available after .step() call)
2) Calling `del p.grad_sample` and `p.summed_grad` after they've been used, the way it's implemented in the old opacus
  a) Code hard to follow (I always forget where grad_samples are cleared in the current version)
  b) Better than the first option, but debugging access is still limited: user has access to `p.grad`, but not 'p.grad_sample'. 
  c) Better than the first option, but it's somewhat unexpected that NOT calling `zero_grad` has no effect. (Essentially we still do zero_grad for the user, but implicitly)

Instead, I propose to detect situations when user didn't call `zero_grad` and throw an exception forcing them to do so.

### Other
- Added test for `make_private_with_epsilon`, fixed the method call. Also print warning if called on a non-empty accountant.
- Validate and throw a user-friendly message if client uses Poisson sampling with IterableDataset
- renamed `DPOptimizer.optimizer` to `DPOptimizer.original_optimizer` (to avoid things like `optimizer = optimizer.optimizer`)
- Check if dataset has changed within one PrivacyEngine (i.e. one accountant), log warning if yes. There're legitimate cases for this, but there's no(?) way to do a robust check
- Deleted old clipping code


opacus/optimizers/optimizer.py
```diff
@@ -1,12 +1,37 @@
...
+from typing import Callable, List, Optional, Union
 
 import torch
 from torch import nn
 from torch.optim import Optimizer
 
 
+def _label_processed(obj: Union[torch.Tensor, List[torch.Tensor]]):
+    if isinstance(obj, torch.Tensor):
+        obj.processed = True
```
karthikprasad, line 12: nit: make this private attribute and more explicit. `_opacus_processed` ?

opacus/optimizers/optimizer.py
```diff
@@ -1,12 +1,37 @@
...
 
-from typing import Callable, List, Optional
+from typing import Callable, List, Optional, Union
 
 import torch
 from torch import nn
 from torch.optim import Optimizer
 
 
+def _label_processed(obj: Union[torch.Tensor, List[torch.Tensor]]):
```
karthikprasad, line 10: nit: `_mark_as_processed` ? 
I read this as the class label was processed in the first reading.

opacus/privacy_engine.py
```diff
@@ -53,8 +53,11 @@ def __init__(self, accountant: str = "rdp", secure_mode=False):
 
         """
         self.accountant = get_accountant(mechanism=accountant)
+        self.accounting_mechanism = accountant
```
karthikprasad, line 56: Suggestion: we can have an attribute in accountant and then access it as `self.accountant.mechanism`

opacus/privacy_engine.py
```diff
@@ -53,8 +53,11 @@ def __init__(self, accountant: str = "rdp", secure_mode=False):
 
         """
         self.accountant = get_accountant(mechanism=accountant)
```
karthikprasad, line 55: I do like the arg in constructor being called just `accountant`, but we use `accountant_mechanism` elsewhere. Do you think it is useful to make this consistent?

opacus/privacy_engine.py
```diff
@@ -132,6 +146,17 @@ def _prepare_model(
...
+                module.batch_first != batch_first
+                or module.loss_reduction != loss_reduction
+            ):
+                raise ValueError(
+                    f"Pre-existing GradSampleModule doesn't match new arguments."
+                    f"Got: module.batch_first: {module.batch_first}, module.loss_reduction: {module.loss_reduction}"
+                    f"Requested: batch_first:{batch_first}, loss_reduction: {loss_reduction}. "
+                    f"Please pass vanilla nn.Module instead"
+                )
+
```
karthikprasad, line 159: I propose moving this inside validator. Validator's gotta validate! ;)

opacus/privacy_engine.py
```diff
@@ -239,25 +262,28 @@ def make_private_with_epsilon(
...
-        sigma_min: Optional[float] = None,
-        sigma_max: Optional[float] = None,
+        **kwargs,
     ):
         sample_rate = 1 / len(data_loader)
 
+        if len(self.accountant) > 0:
+            warnings.warn(
+                "You're calling make_private_with_epsilon with non-zero privacy budget "
+                "already spent. Returned noise_multiplier assumes zero starting poing, "
```
karthikprasad, line 272: typo

karthikprasad (commented):
As always, love the detailed summary. And especially love
> Instead, I propose to detect situations when user didn't call zero_grad and throw an exception forcing them to do so.

opacus/optimizers/optimizer.py
ffuuugor: + on making private

On "opacus_processed" vs "processed" - do you think renaming make sense even if we only attach this to opacus-owned attributes (namely `p.grad_sample` and `p.summed_grad`)

karthikprasad: probably not for grad_sample since it is not necessarily for opacus alone. It is an attribute of module after all.

On the other hand, processed was solely for use by opacus? But then again, one could argue it is an attribute of the module irrespective of who uses it. In conclusion, it is a weak nit, and I am happy to withdraw it :)

opacus/privacy_engine.py
ffuuugor: Is there a missing "not"? If yes, then I agree :)

karthikprasad: No, there is no missing "not" :)

ffuuugor: Oh, I see what you mean
You like both consistency and args being called "accountant"

The inconsistency I wanted to address here is that some `accountant` parameters/variables are strings, others are actual accountant objects. Are we ok with this?

opacus/privacy_engine.py
ffuuugor: I'm not sure it's appropriate use of a validator. It's only triggered on a second subsequent call of `make_private`, I'm not sure how validator could check things based on a context

opacus/privacy_engine.py
karthikprasad: Yeah. Consistency would mean changing the arg name in PrivacyEngine constructor to `accountant_mechanism`. that way, all accountant objects will be `accountant` and strings will be `accountant_mechanism`

opacus/accountants/utils.py
```diff
@@ -0,0 +1,46 @@
...
+    target_delta: float,
+    sample_rate: float,
+    epochs: int,
+    accountant: str = "rdp",
+    **kwargs,
+) -> float:
+    r"""
+    Computes the noise level sigma to reach a total budget of (target_epsilon, target_delta)
+    at the end of epochs, with a given sample_rate
+    Args:
```
romovpa, line 15: Extra lines before `Args:` and `Returns:` are probably required to render docs correctly.

opacus/accountants/utils.py
```diff
@@ -0,0 +1,46 @@
...
+        sample_rate: the sampling rate (usually batch_size / n_data)
+        epochs: the number of epochs to run
+        alphas: the list of orders at which to compute RDP
+    Returns:
+        The noise level sigma to ensure privacy budget of (target_epsilon, target_delta)
+    """
+    eps = float("inf")
+    sigma_min = 0.01
+    sigma_max = 10.0
+    accountant = get_accountant(mechanism=accountant)
```
romovpa, line 27: Missed the change where you introduced this :(

First, we need to rename this to something like `create_accountant`. Otherwise it looks like there is one global accountant for the mechanism name string and you tries to retrieve it.

I understand using an optional str value in the accountant field instead of `IAccountant` instance. But why use it here? 
Can accountant of some type have hyperparameters?

opacus/accountants/utils.py
```diff
@@ -0,0 +1,46 @@
...
+        target_delta: the privacy budget's delta
+        sample_rate: the sampling rate (usually batch_size / n_data)
+        epochs: the number of epochs to run
+        alphas: the list of orders at which to compute RDP
+    Returns:
+        The noise level sigma to ensure privacy budget of (target_epsilon, target_delta)
+    """
+    eps = float("inf")
+    sigma_min = 0.01
+    sigma_max = 10.0
```
romovpa, line 26: Why these magic constants?

opacus/accountants/utils.py
```diff
@@ -0,0 +1,46 @@
...
+    accountant = get_accountant(mechanism=accountant)
+
+    while eps > target_epsilon:
+        sigma_max = 2 * sigma_max
+        accountant.steps = [(sigma_max, sample_rate, int(epochs / sample_rate))]
+        eps = accountant.get_epsilon(delta=target_delta, **kwargs)
+        if sigma_max > 2000:
+            raise ValueError("The privacy budget is too low.")
+
+    while sigma_max - sigma_min > 0.01:
```
romovpa, line 36: Two more magic constants. Maybe add comments to them?

opacus/optimizers/optimizer.py
```diff
@@ -1,13 +1,38 @@
...
+            x._processed = True
+
+
+def _check_processed_flag_tensor(x: torch.Tensor):
+    if hasattr(x, "_processed"):
+        raise ValueError(
+            "Gradients hasn't been cleared since the last optimizer step. "
+            "In order to obtain privacy guarantees you must call optimizer.zero_grad()"
+            "on each step"
+        )
```
romovpa, line 25: Check my understanding please: `zero_grad` not only changes `.grad`, but also clears all the subfields? Otherwise, how it works? In [the docs](https://pytorch.org/docs/stable/generated/torch.optim.Optimizer.zero_grad.html) it is said that `zero_grad` only changes `.grad`.

opacus/optimizers/optimizer.py
romovpa: I vote for `_opacus_processed` and request short explanation in the docstring (why we need this label and why it is removed after zero_grad).

opacus/accountants/accountant.py
```diff
@@ -15,5 +15,10 @@ def get_epsilon(self, delta: float, *args, **kwargs) -> float:
         pass
 
     @abc.abstractmethod
-    def __len__(self):
+    def __len__(self) -> int:
+        pass
+
+    @classmethod
```
karthikprasad, line 21: nit: also add `@property`?

opacus/optimizers/optimizer.py
romovpa: Wait, we add this field to Tensor, not to tensor.grad_sample, right?

romovpa (commented):
>  I propose to detect situations when user didn't call zero_grad and throw an exception forcing them to do so.

I like this third option.

opacus/privacy_engine.py
karthikprasad: > I'm not sure it's appropriate use of a validator.

It does not belong in `ModuleValidator`, but does belong in `self.validator()`?

> I'm not sure how validator could check things based on a context

We are already passing data_loader and optimizer to self.validator() even though they aren't being used. We can supply the context too.

opacus/accountants/accountant.py
romovpa: Would be good, but folks say this [only works only in py3.9+](https://stackoverflow.com/a/1800999)

karthikprasad: Ugh! OK. Oh well!

opacus/accountants/utils.py
ffuuugor: Let's deal with the docstrings separately

opacus/accountants/utils.py
ffuuugor: agree on `create_accountant` naming.

We don't want to pass `accountant` object from PrivacyEngine to here, because this method messes with accountant's internal state (changing accountant.steps) to do a binary search for an appropriate noise multiplier.

We don't want util method to inadvertently change the privacy state of a PrivacyEngine, so it's safer to create new object instead

opacus/optimizers/optimizer.py
ffuuugor: we override `zero_grad` method in DPOptimizer and GradSampleModule to also clear fields created and managed by opacus (i.e. `grad_sample` and `summed_grad`)

opacus/privacy_engine.py
ffuuugor: It's a toughie - I also like args being called `accountant`.
Does consistency here overweights clear arg name for you? After all, consistency is for the internal implementation, but constructor arg is for everyone

opacus/optimizers/optimizer.py
ffuuugor: Nope, we add it to `p.grad_sample` (or, in case of a multiple accumulated batches, where `.grad_sample` is a list - each tensor in that list)

it's `p.grad_sample._processed` and `p.summed_grad._processed`, never `p._processed`

opacus/data_loader.py
```diff
@@ -87,12 +87,8 @@ def __init__(
     def from_data_loader(
         cls, data_loader: DataLoader, distributed: bool = False, generator=None
     ):
-        if isinstance(data_loader, cls):
-            # TODO: this should be exception, not assert
-            assert data_loader.distributed == distributed
```
alexandresablayrolles, line 92: Why was this removed?

alexandresablayrolles (commented):
I'm not a big fan of adding the complexity of "processed" to the code. The only downside to option 2) seems to be the complexity but for me this is the situation:
- we cannot do gradient accumulation in .grad anymore because Pytorch puts some non-private info there
- however we *can* do gradient accumulation in .summed_grad and it's "private" accumulation (i.e. we accumulate before adding noise - even though with the gaussian mechanism and RDP it does not make any difference)
- grad_sample is "consumed" by clip_and_accumulate, it's the logical thing to have

ffuuugor:
> The only downside to option 2) seems to be the complexity but for me this is the situation

I'd say for the the biggest selling point for option 3 is transparency - clients will always know whether the gradients are cleared or not.
I think minimising surprise factor is a good design principle, and option 3 wins over option 2 in this regard.

opacus/data_loader.py
ffuuugor: What's the upside of having it? Looks like we can easily re-create DPDataLoader from another DPDataLoader and change distributed flag, so why not do it

opacus/privacy_engine.py
ffuuugor: Are you sure you don't like it the way it's implemented now? It's always `accountant: str` when it's a parameter and it's always `self.accountant: IAccountant` when it's an attributed.
all `*_mechanism` is moved to accountant object as you suggested

opacus/optimizers/optimizer.py
karthikprasad: > it's p.grad_sample._processed and p.summed_grad._processed, never p._processed

Oh yes! I totally missed this. In that case, I completely withdraw my nit. For some reason, I was under the impression that it will be `p._processed`.  `grad_sample` is created by us, so `p.grad_sample._processed` is more than sufficient.

opacus/privacy_engine.py
karthikprasad: Sorry, I missed replying yesterday. 

> It's a toughie - I also like args being called accountant.

+1

> Does consistency here overweights clear arg name for you? 

Doesn't overweigh, probably equal. Hence the equivocality. :/

> Are you sure you don't like it the way it's implemented now?

I think I can live with it. Should it ever get confusing, I would propose changing `self.accountant` to `self.accountant_obj` or some such thing. I like `accountant: str` as an arg. If no one else feels strongly, I vote for keeping things as is for now.

romovpa (approved):
LGTM! 👍

PR #265, opened 2021-11-17
Unit tests when batch_first=False for embedding and linear layers

ashkan-software:
## Types of changes

Looks like our current tests do not cover when `batch_first=False`. This PR adds those tests for linear and embedding layers. However, when the input dimension is 2 for linear and 1 for embedding, the tests fail. I added that as todo, as it seems like our helper functions may a few bugs

I ran the following tests locally:
```
python -m unittest opacus/tests/grad_samples/linear_test.py
python -m unittest opacus/tests/grad_samples/embedding_test.py
```

## Checklist


opacus/tests/grad_samples/common.py
```diff
@@ -200,22 +200,16 @@ def run_test(
...
-            module,
-            batch_first=batch_first,
-            loss_reduction="sum",
-            atol=atol,
-            rtol=rtol,
-        )
+        for loss_reduction in ["mean", "sum"]:
+            for batch_first in [True, False]:
+                self.run_test_with_reduction(
+                    x,
```
romovpa, line 206: Should we change the input for the different `batch_first` settings?

ashkan-software:
Thanks @romovpa for the review. I am still working on this and just put it up so that the tests run :)

opacus/tests/grad_samples/dp_rnn_test.py
```diff
@@ -76,8 +76,5 @@ def test_rnn(
         if using_packed_sequences:
             x = _gen_packed_data(N, T, D, batch_first, packed_sequences_sorted)
         else:
-            if batch_first:
-                x = torch.randn([N, T, D])
-            else:
-                x = torch.randn([T, N, D])
+            x = torch.randn([N, T, D])
```
ashkan-software, line 79: looks like the if-else condition is not useful, as we already swap the input in https://github.com/pytorch/opacus/blob/3f3cc88340b2476cf685a3c2eb1f03b70ccfff9b/opacus/tests/grad_samples/common.py#L122

ffuuugor: That's incorrect.

Input data should always match `batch_first` flag.

a) Line you quote only works in `compute_microbatch_grad_sample` branch - that's half of the test (reminder: the test runs `compute_microbatch_grad_sample` and `compute_opacus_grad_sample` to make sure they produce identical results.)

b) If you look further down to lines 128-130 https://github.com/pytorch/opacus/blob/3f3cc88340b2476cf685a3c2eb1f03b70ccfff9b/opacus/tests/grad_samples/common.py#L128-L130

we then put batch dimension back, while respecting `batch_first` flag

opacus/tests/grad_samples/linear_test.py
```diff
@@ -27,9 +28,10 @@ def test_input_bias(
         H: int,
         input_dim: int,
         bias: bool,
+        batch_first: bool,
     ):
 
-        if input_dim == 2:
+        if input_dim == 2:  # TODO: fix when dim is 2
```
ffuuugor, line 34: I don't understand: I think `batch_first` flag should have an effect on `x_shape`

ffuuugor (changes requested):

opacus/tests/grad_samples/dp_rnn_test.py
ashkan-software: Hmm good point. Do you have any idea why the tests pass then? Is it because the input is just randomly generated?

opacus/tests/grad_samples/linear_test.py
ashkan-software: So I just changed the tests (line 18) to check `input_dim = 3` and `4`.

opacus/tests/grad_samples/dp_rnn_test.py
ashkan-software: I still don't understand why you think this change is not correct. We are changing the `x` on line 122 of common.py, since you wrote

> Input data should always match batch_first flag.

ashkan-software:
I let the if-else statement be as is @ffuuugor for faster review and merge of this. We can discuss it while this PR is being merge

opacus/tests/grad_samples/linear_test.py
```diff
@@ -27,15 +28,16 @@ def test_input_bias(
...
         if input_dim == 3:
             x_shape = [N, Z, W]
         if input_dim == 4:
             x_shape = [N, Z, H, W]
 
         linear = nn.Linear(W, W + 2, bias=bias)
         x = torch.randn(x_shape)
-        self.run_test(x, linear, batch_first=True)
+        if not batch_first:
+            x = x.transpose(0, 1)
```
ashkan-software, line 42: FYI: I also added the transpose in this test, even though the test passes it without it. The test passes regardless of this line, as the input is just bunch of random numbers and the 0'th and fist dimension do not matter in this test

ffuuugor: Ok, I see what you mean - for Linear and Embedding the first dimensions doesn't matter at all, so when we have `batch_first=False` and `x_shape = [N, Z, W]`, then batch dimension is `Z` not `N`. This is what confused me as a reader.

ffuuugor (approved):
Thanks! And sorry for unnecessary back and forth.

Before merging, can you pls bring back `input_dim=2` case, but only with `batch_first=True`:
```
if input_dim == 2:
  if not batch_first:
    return
  else:
    # do stuff
```

PR #266, opened 2021-11-18
Parameterize grad sample registry

karthikprasad:
## Types of changes

The current grad sampler registrar was hard coded to Opacus supplied GradSampleModule. This PR removes the hard code and allows anyone to pass their own GradSampleModule with different registries.
This follows the same pattern as that of validator registrar.

Unit tests. Backward compatible change.

## Checklist


romovpa (commented):
Could you please give more context, what is the use case?
Why do we need to create a custom GradSampleModule? How to use a custom GradSampleModule? Should we make it changeable in PrivacyEngine as well?

karthikprasad:
> Could you please give more context, what is the use case? Why do we need to create a custom GradSampleModule? How to use a custom GradSampleModule?

Primarily, this PR mimics the pattern of ModuleValidator. The rationale is as follows:
Consider a case where a user wants to create a custom `PrivacyEngine` with custom `GradSampleModule`. Eg: use the fancy conv approach instead of the unfold approach. If they were to do that, they will have to make changes to opacus code as currently, the `register_grad_sampler` decorator in the utils registers the grad_samplers to `GradSampleModule`.

With this change, they won't have to make changes to Opacus code. They can subclass `PrivacyEngine`, use `CustomGradSampleModule` with it, and still continue to use the same `register_grad_sampler` decorator by passing `CustomGradSampleModule` as an arg.

> Should we make it changeable in PrivacyEngine as well?

No. This doesn't change our offered `PrivacyEngine`. We use our `GradSampleModule` and `ModuleValidator`. This only applies to subclassed custom Engines.

alexandresablayrolles:
Shouldn't we have only one "GradSampleModule" class, but each object can chose to have different grad_sampler implementations? We could add to the GSM object an object attribute "grad_samplers" that would have higher priority over the class one.

karthikprasad:
> Shouldn't we have only one "GradSampleModule" class, but each object can chose to have different grad_sampler implementations?

Not necessarily; users can subclass GradSampleModule if they want to.

> We could add to the GSM object an object attribute "grad_samplers" that would have higher priority over the class one.

I am not sure I understand this correctly. From what I gather, your suggestion would result in maintaining a list of grad_samplers in the GSM. This was the original design which we wanted to move away from and hence introduced this `register` decorator.

ffuuugor:
>Shouldn't we have only one "GradSampleModule" class, but each object can chose to have different grad_sampler implementations?

You can actually achieve that with the current API. register method overrides previous grad_sampler, so if you want to use specific one - just call register method explicitly before using GradSampleModule

romovpa:
@karthikprasad This PR stalled. Do you still think we need this change?

Your motivation sounds reasonable to me. 
At the same time @ffuuugor claims with v1.0 we can go without this change.

I would recommend to make this PR a draft or close until we have a specific example where this can be useful, or you can introduce one in this PR. I think the lack of examples doesn't allow us to discuss substantially.

PR #267, opened 2021-11-18
address TODOs, part 2

ffuuugor:
Second and final part of addressing TODOs in the codebase. What's left there can be fixed later.

### More changes to `zero_grad`

We're bringing the empty grad behaviour closer to PyTorch.
Currently if you have empty gradients, `p.grad` would be either None or tensor of zeros, while `p.grad_sample` would throw an `AttributeError`.
With this change we add `grad_sample` attribute to all `GradSampleModule` parameters, setting it to None initially.

We still can't support setting empty grads to tensor of 0s, because accumulation rules are different between regular grads and per-sample grads.

### Noise_seed -> Noise generator

As I found out, there's no (easy?) way to move an existing `torch.Generator` between devices. 
So we'd have to let clients deal with it. Current implementation doesn't let client provide a device with which to initialize Generator. So instead we'd ask them to provide generator iteself

### Other

- Removed reference to mypy. We've had a todo item to bring it back, but since then we're increasingly rely on monkeypatching, so I don't think we'll be able ti bring it back any time soon
- GDP accountant: we've had a no-op if-else branch. cc @alexandresablayrolles to check if changes to `gdp.py` make sense
- Wrapped couple more methods from `torch.optim.Optimizer` for `DPOptimizer`. Didn't test whether it works, but looks ok

### DIFF
--- a/opacus/grad_sample/grad_sample_module.py
+++ b/opacus/grad_sample/grad_sample_module.py
@@ -46,7 +46,7 @@ def create_or_accumulate_grad_sample(
 
 def promote_current_grad_sample(p: nn.Parameter) -> None:
     if p.requires_grad:
-        if hasattr(p, "grad_sample"):
+        if p.grad_sample is not None:
             if isinstance(p.grad_sample, list):
                 p.grad_sample.append(p._current_grad_sample)
             else:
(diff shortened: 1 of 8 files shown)


opacus/grad_sample/grad_sample_module.py
```diff
@@ -80,56 +80,60 @@ def __init__(
...
-                else:
-                    grad_samples = [p.grad_sample]
+            p.grad_sample = None
 
-                for grad_sample in grad_samples:
-                    if grad_sample.grad_fn is not None:
-                        grad_sample.detach_()
-                    else:
-                        grad_sample.requires_grad_(False)
+    def del_grad_sample(self):
```
karthikprasad, line 126: Who uses this now?

karthikprasad (approved):
Looks clean! Thanks a lot.

opacus/grad_sample/grad_sample_module.py
ffuuugor: `to_standard_module` method - we still want to delete if we're doing full detach from the module

opacus/grad_sample/grad_sample_module.py
```diff
@@ -318,7 +322,6 @@ def rearrange_grad_samples(
                 " run forward after add_hooks(model)"
             )
 
-        # TODO: that's not right; DPLSTM isn't always batch_first=False
         batch_dim = 0 if batch_first or type(module) is RNNLinear else 1
```
romovpa, line 325: I just found that this line does something hacky. A don't understand why we propagate batch_first to all submodules in `add_hooks()`, is this what we want to do?

Why `RNNLinear` is the only exception here? Does this work correctly with `nn.Linear`? I have a feeling that this is a bug not covered by tests.

The method `add_hooks()` goes recursively through all submodules and set the global `batch_first`:
```python
        for module in trainable_modules(self._module):
            if type(module) in self.GRAD_SAMPLERS:
                ...
                self.autograd_grad_sample_hooks.append(
                    module.register_backward_hook(
                        partial(
                            self.capture_backprops_hook,
                            ...
                            batch_first=batch_first,
                        )
                    )
                )
```

I found this behavious counter-intuitive because the module can easily have submodules with different batch_dim and dim rearranges in between. For example, in `DPRNN`:
```
model: DPRNN(batch_dim=1)
  - cell: nn.Linear(batch_dim=0)
```
In case of DPRNN (as well as the original nn.RNN) batch rearranging happens inside the implementation. 

If not that `or type(module) is RNNLinear` we would try to treat nn.Linear as it uses `batch_dim=1` which is not the case, it alwayse uses batch_dim=0.

ffuuugor: Based on what you described, I don't think we have a clear bug. 
But we certainly have an implicit assumption about the model, which we should communicate better. (If you have any ideas on how to verify / enforce it, they'll be very welcome)

In addition to everything we check in `ModelValidator`, we also assume the model doesn't do any weird reshapes. On all intermediate stages we expect the model to maintain batch_dimension and either first or second dimension. (and yes, we don't verify that). 
On a high level, it's the same as us prohibiting `BatchNorm` - samples must be processed independently of each other, and the output of any batch must be equivalent to micro-brach processing (process each sample individually with batch_size=1 and average the results). It's not the case for `BatchNorm`, and it's also not the case if you do custom tensor operations and don't preserve batch dimension.

> A don't understand why we propagate batch_first to all submodules in add_hooks(), is this what we want to do?

In the current implementation - yes, because of the implicit assumption above.

> Why RNNLinear is the only exception here? Does this work correctly with nn.Linear?

Because we assume `RNNLinear` to be only used in `DPRNN`, and we know `DPRNN` does this with batch dimension.
It's also the reason we can't get rid if `RNNLinear` in favour of `nn.Linear` just yet

romovpa: Thanks for explaining me. Wow, this is a strong assumption, it's definitely worth to make it clear.

Although, I'm still trying to understand how it can work.

> On all intermediate stages we expect the model to maintain batch_dimension and either first or second dimension. (and yes, we don't verify that).

What happens to the inner nn.Linear when we use batch_dim=1? 
This hack with RNNLinear looks to me as a solution of this problem, but how other modules work?
For example, in `DPMultiheadAttention` is always batch_dim=1, but there are nn.Linear layers which always batch_dim=0.

PR #268, opened 2021-11-19
Purge stats.py and utils/clipping.py (Unused code)

ffuuugor:
Removing some unused code.

`opacus/utils/clipping.py` is straightforward, I just forgot to remove it along with `per_sample_gradient_clip.py` in the previous PR.

`opacus/utils/stats.py` is different. I'm not sure if it's actually dead code or I just don't understand it's applications. Does anyone remember the context around it?


karthikprasad:
`opacus/utils/stats.py` was added to capture stats for different clipping strategies for slicing and dicing on TB

ffuuugor:
Does it have any good use cases with the current architecture?
Looks to me like a tb wrapper, heavily tuned for `ClippingStrategy` enum - so my guess would be no?

karthikprasad:
> Does it have any good use cases with the current architecture? Looks to me like a tb wrapper, heavily tuned for `ClippingStrategy` enum - so my guess would be no?

Yup. Purging should be fine. May be change the title of this PR to explicitly state that stats collector is deleted so that we can easily find it later if we have to.

karthikprasad (approved):

alexandresablayrolles:
In the future this will definitely belong to Opacus-lab! Looks good to me

alexandresablayrolles (approved):

PR #272, opened 2021-11-23
[RFC] Add debug mode

alexandresablayrolles:
This is an RFC for a "debug mode" for Opacus. 
Often when using DP, things do not work as expected and it can be useful to have access to some lower-level signals. In particular, gradient norms are computed by the DPOptimizer but are then "thrown away".
The idea of the debug mode is to typically log this information to make it accessible to the user.

### DIFF
--- a/opacus/optimizers/optimizer.py
+++ b/opacus/optimizers/optimizer.py
@@ -97,6 +97,7 @@ class DPOptimizer(Optimizer):
         loss_reduction: str = "mean",
         generator=None,
         secure_mode=False,
+        debug=False,
     ):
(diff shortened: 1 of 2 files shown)


ffuuugor:
As discussed offline, I'd suggest refactoring this proposal as an independent util-method.
We don't need to additionally store grad_norms, as we already have access to raw per sample gradients. Cleaner solution would be not to store duplicate information, but rathe provide an easy way to compute norms based on stored per sample gradients

alexandresablayrolles:
Given that it's a two-line function I'm not sure it's very useful inside utils. Abandoning this diff.

PR #273, opened 2021-11-23
1.0 API

ffuuugor:
We're introducing new Opacus API. See Readme and updated turorials for details.

For code reviews and discussions see PR history https://github.com/pytorch/opacus/pulls?q=is%3Apr+is%3Aclosed+base%3Aexperimental_v1.0

PR #277, opened 2021-11-24
require python 3.7 + minor refactoring

ffuuugor:
Following some internal test failures on #273, we've decided to up min required python version (we use annotations from __future__, which are only introduced in 3.7). Since 3.7 is already a default in colab, and is pretty popular overall, I don't see many potential issues with it.

One more change (which probably should have been decoupled into a separate pull request) is accountant hook refactoring. While writing guide on advanced usage of opacus, I realised that accountant hook is always the same for a given accountant. It can be different for different accountants, but same mechanism always requires the same hook. 

This change makes it easier to attach accountant outside of PrivacyEngine

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,15 +7,15 @@ version: 2.1
 commands:
 
   py_3_7_setup:
-    description: "Install and switch to Python 3.7.10; also install pip and pytest."
+    description: "Install and switch to Python 3.7.4; also install pip and pytest."
     steps:
       - run:
-          name: "Setup Python v3.7.10 environment"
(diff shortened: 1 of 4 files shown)


ffuuugor:
From offline discussion with @karthikprasad :

> a) All things equal we don't want to add unnecessary restrictions. We want to only set requirements that are based on which features we're using
b) We should only support versions we can regularly test. if we can't set up regular testing with a certain python version, it'll be too easy to slip and introduce breaking change

With that, we're choosing the oldest python 3.7 version available on circleCI: 3.7.4

karthikprasad (approved):

PR #278, opened 2021-11-24
Add perlayer multigpu test

alexandresablayrolles:
Include per-layer clipping to the multi-gpu test. Remove validator that filters out DDP


opacus/tests/multigpu_gradcheck.py
```diff
@@ -67,21 +67,25 @@ def demo_basic(rank, weight, world_size, dp):
...
         ddp_model = DPDDP(model)
     else:
         ddp_model = DDP(model, device_ids=[rank])
 
     privacy_engine = PrivacyEngine()
 
     if dp:
+        max_grad_norm = 1e8
+        if clipping == "per_layer":
+            max_grad_norm = [1e8 for p in model.parameters()]
```
karthikprasad, line 80: Why not use `inf` instead?

ffuuugor (approved):
LGTM

I think in the future we might think how to bring ddp check back (since it's only allowed with per layer optimizer). I think it's doable within privacy engine, but out of scope for this PR

PR #280, opened 2021-11-24
Fix distributed batch size

alexandresablayrolles:
Fix distributed batch size


opacus/privacy_engine.py
```diff
@@ -334,6 +334,8 @@ def make_private(
 
         sample_rate = 1 / len(data_loader)
         expected_batch_size = int(len(data_loader.dataset) * sample_rate)
+        if distributed:
+            expected_batch_size /= torch.distributed.get_world_size()
```
ffuuugor, line 338: do you think it's better to put it to DistributedDPOptimizer / DistributedPerLayerOptimizer `__init__` method??

ffuuugor:
This we should abandon, right?

PR #288, opened 2021-11-26
Advanced tutorial

ffuuugor:
New tutorial, that introduced user to `GradSampleModule`, `DPOptimizer` and `DPDataLoader`. 

It unpacks the inner workings of `PrivacyEngine.make_private()` to explain what it does and enable people to use only parts of it, if the `PrivacyEngine` API feels too restrictive for their use case.

We also take a look at `DPOptimizer` implementation and learn how to subclass it


ashkan-software:
Great work! 

Any plan on putting it here? https://opacus.ai/tutorials/

PR #296, opened 2021-11-30
add guide to grad sampler tutorial

karthikprasad:
Summary: Adding a tutorial explaining grad sampler registration and GradSampleModule.

Differential Revision: D32734269

### DIFF
--- a/tutorials/guide_to_grad_sampler.ipynb
+++ b/tutorials/guide_to_grad_sampler.ipynb
@@ -1,0 +1,272 @@
+{
+  "nbformat": 4,
+  "nbformat_minor": 0,
+  "metadata":{
+    "kernelspec":{
+      "name":"python3",
+      "display_name":"Python 3 (ipykernel)",
+      "language":"python"
+    },
+    "language_info":{
+      "codemirror_mode":{
+        "name":"ipython",
+        "version":3
+      },
+      "file_extension":".py",
+      "mimetype":"text/x-python",
+      "name":"python",
+      "nbconvert_exporter":"python",
+      "pygments_lexer":"ipython3",
+      "version":"3.7.5"
+    }
+  },
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "vfch6fQlnG05"
+      },
+      "source": [
+        "# Guide to grad samplers"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "RKnAlq0AnQCd"
+      },
+      "source": [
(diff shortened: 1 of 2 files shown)


tutorials/guide_to_grad_sampler.ipynb
```diff
@@ -0,0 +1,241 @@
...
+        "# Guide to grad samplers"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "RKnAlq0AnQCd"
+      },
+      "source": [
+        "DP-SGD guarantees privacy of every sample used in the training. In order to realize this, we have to bound the sensitivity of every sample, and in order to that we have to clip the gradient of every sample. Unfortunately, pytorch doesn't maintain the gradients of individual samples in a batch and only exposes the aggregated gradients of all the samples in a batch via the [`.grad` attribute](https://pytorch.org/docs/stable/generated/torch.Tensor.grad.html). "
```
ffuuugor, line 39:
```suggestion
        "DP-SGD guarantees privacy of every sample used in the training. In order to realize this, we have to bound the sensitivity of every sample, and in order to do that we have to clip the gradient of every sample. Unfortunately, pytorch doesn't maintain the gradients of individual samples in a batch and only exposes the aggregated gradients of all the samples in a batch via the [`.grad` attribute](https://pytorch.org/docs/stable/generated/torch.Tensor.grad.html). "
```

tutorials/guide_to_grad_sampler.ipynb
```diff
@@ -0,0 +1,241 @@
...
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "ERr4O-dVXMne"
+      },
+      "source": [
+        "\n",
+        "The easiest way to get what we want is to train with batch size of 1 as follows:\n"
```
ffuuugor, line 49: This is missing a cell

ffuuugor (approved):
Lovely, thanks! Links in the last paragraph are a great addition!

Just as an idea (we don't have to do it now, but if we like it, we can create an issue for the future), we should also explain our grad sampler tests - people can use the same utils to check if their custom grad sampler is correct.

Also, we cover math behind vectorized grad samplers in this post https://medium.com/pytorch/differential-privacy-series-part-2-efficient-per-sample-gradient-computation-in-opacus-5bf4031d9e22, probably worth linking to

tutorials/guide_to_grad_sampler.ipynb
karthikprasad: Great catch!

PR #299, opened 2021-12-02
fix docstring for PrivacyEngine.get_epsilon()

jessijzhao:
Fix docstring for `PrivacyEngine.get_epsilon()`

While `RDP.get_privacy_spent(delta, alphas)` returns both epsilon and optimal alpha,`self.accountant.get_epsilon(delta)` returns only epsilon for both GDP and RDP.


ffuuugor (approved):
Good catch, thanks!

Issue #322, opened 2022-01-03
DPOptimizer should pass through defaults


gkaissis:
DPOptimizer doesn't pass through the .defaults attribute of the original optimizer. Since this can include things like momentum, this can break code using schedulers. For example, when using NAdam, schedulers with cycle_momentum=True will not fail as intended because NAdam has no momentum, but will instead fail on the attribute lookup for the .defaults attribute on the wrapped optimizer.
Proposed fix:

self.defaults = self.original_optimizer.defaults

alexandresablayrolles:
Thanks for flagging, we have a PR to fix it (https://github.com/pytorch/opacus/pull/329)

alexandresablayrolles:
Closing this issue as PR was merged.

PR #329, opened 2022-01-18
Fix: Passing through `defaults` in DPOptimizer

alexandresablayrolles:
## Problem 

Described in #322. The code using schedules can be broken because DPOptimizer doesn't have `.defaults` field. For example, when using NAdam, schedulers with cycle_momentum=True.

## Solution

As suggested by @gkaissis, just passing through the `defaults` field to DPOptimizer.

Differential Revision: D33634214


romovpa:
@alexandresablayrolles Could you please add more comments: why we need to match `defaults`? What is the usage of this?
Could there be other fields we need to match?

romovpa (approved):

Issue #313, opened 2021-12-16
The Lightning example doesn't train with privacy?


tholop:
## 🐛 Bug
I think that there is something strange with the PyTorch Lightning example. First, I realized that changing `max_grad_norm` had no impact on the accuracy. Then, using `noise_multiplier = 1000` gave much higher accuracy than expected (see below).

In the code, it seems that the DP version of the model is stored somewhere but never used:
https://github.com/pytorch/opacus/blob/eedb4447b4b5271591c268c6f280823b2415e134/examples/mnist_lightning.py#L107

Unless there is something subtle going on (maybe you fetch `self.dp` elsewhere in the optimizer when you detect a Ligthning module?), the forward pass is using the non-DP model instead of `self.dp["model"]`:
https://github.com/pytorch/opacus/blob/eedb4447b4b5271591c268c6f280823b2415e134/examples/mnist_lightning.py#L113

Please, did I miss anything?

## To Reproduce
Download `mnist_lightning.py` and run the following:
```
❯ python mnist_lightning.py fit --model.max_grad_norm 1 --model.noise_multiplier 1000  --trainer.max_epochs 3                           
GPU available: True, used: False                                                                                                              
TPU available: False, using: 0 TPU cores                                                                                                      
IPU available: False, using: 0 IPUs                                                                                                           
Epoch 2: 100%|████████████████████████████████████| 1500/1500 [00:27<00:00, 54.03it/s, loss=0.0488, v_num=2, train_loss=0.0685, epsilon=0.103]
Epoch 2: 100%|██████████████████████████████████████████████████████████████████████| 1500/1500 [00:25<00:00, 58.78it/s, loss=0.0377, v_num=2]
Restoring states from the checkpoint path at /home/.../lightning_logs/version_2/checkpoints/epoch=2-step=5999.ckpt                   
Loaded model weights from checkpoint at /home/.../lightning_logs/version_2/checkpoints/epoch=2-step=5999.ckpt                        
Testing:  97%|██████████████████████████████████████████████████████████████████████████████████████████▎  | 304/313 [00:01<00:00, 155.37it/s]--------------------------------------------------------------------------------
DATALOADER:0 TEST RESULTS                                                                                                                     
{'test_accuracy': 0.9854999780654907, 'test_loss': 0.04740014299750328}                                                                       
--------------------------------------------------------------------------------
```

romovpa:
@tholop Thank you for reporting this!

[...]

PR #330, opened 2022-01-19
Fix: Closure call after applying DP noise

romovpa:
## Types of changes

## Bug

In DPOptimizer step we apply noise to the gradients after then call `original_optimizer.step()`. 
If `closure` is used we should run it manually and then re-apply noise before making an optimizer step. The problem is that we pass closure to the original_optimizer causing it to re-compute gradients and clear up the noise. This removes the DP functionality. 

Closure is used when the optimizer needs more than one forward-backward passes. Also it's always used when using [automatic optimization](https://pytorch-lightning.readthedocs.io/en/stable/common/optimizers.html) in Pytorch Lightning (causing #313).

Interestingly, I already fixed this problem, but we we've lost the change. Probably during me migrating to the v1.0 API or while merging.

## Solution

We solve this by not passing the closure to the original optimizer while running it manually before making a step.

`privacy_engine_test.py` suite is updated to ensure the noise is added when passing `closure` to DPOptimizer.

## Further work

This solution is temporary because it doesn't support optimizers requiring more than one model evaluations to make a step (in this case the optimizer should raise due to `closure=None`). We need a design change to tackle this. What quickly comes to mind, we can wrap closure including add_noise into the wrapper.

Another question: how several model evaluations affect DP. E.g. how DP-LBFGS should work?

### DIFF
--- a/opacus/optimizers/optimizer.py
+++ b/opacus/optimizers/optimizer.py
@@ -476,7 +476,7 @@ class DPOptimizer(Optimizer):
                 closure()
 
         if self.pre_step():
-            return self.original_optimizer.step(closure)
+            return self.original_optimizer.step()
         else:
             return None
 
--- a/opacus/tests/privacy_engine_test.py
+++ b/opacus/tests/privacy_engine_test.py
@@ -131,6 +131,31 @@ class BasePrivacyEngineTest(ABC):
                 if max_steps and steps >= max_steps:
                     break
 
+    def _train_steps_with_closure(
+        self,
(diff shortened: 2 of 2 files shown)


karthikprasad:
Thanks @romovpa. Would it be possible to add a unit test to catch this peculiarity?

romovpa:
> Thanks @romovpa. Would it be possible to add a unit test to catch this peculiarity?

You are right, it's a must. Added the case to the privacy engine test suite.

Issue #308, opened 2021-12-09
Opacus: DP-SGD with massive epsilon does not attain same accuracy as model with DP-SGD


RoyRin:
## ❓ Opacus Training models with DP-SGD with a massive epsilon does not attain same accuracy as model with DP-SGD

I trained Resnet9  with **no DP, and achieved : 85.8 % accuracy**
I then retrained Resnet9 with different epsilon for DP-SGD, and got this curve:
![image](https://user-images.githubusercontent.com/12860243/145493464-ea58a071-0d44-4c52-8f46-66ad83ebf3f9.png)

Whereas **epsilon : 1e106  attains accuracy: 76.7 %**

Note: The code I used to generate this plot is available upon request. And If this is not the right place to raise this issue, please let me know.

ffuuugor:
Thanks for raising - this is the right place, as it indeed looks like a bug and worth looking into for us.

Can you share your input privacy params (noise and max_grad_norm)? Often clipping alone could negatively affect the model - e.g. if you set `max_grad_norm=1.` and `noise_multiplier=1e-7`, you'll get horrendous epsilon, but the model can still underperform compared to the original version

RoyRin:
So originally, I tried it with max_grad_norm = 1.2, but then also re-ran it with max-grad-norm = 1.2*400 

![image](https://user-images.githubusercontent.com/12860243/145692660-1fd11c6d-3c17-45b8-b192-3749f37482fd.png)

Which does even worse...
I didn't explicitly adjust the noise_multiplier, so that could be it.

Here is a gist of my code: I trained this on Resnet9, but also validated it on a few one-off examples of Resnet18 (it just takes longer to train resnet18). https://colab.research.google.com/gist/RoyRin/816282e6e173cb35eb3d1df75687053d/building_image_classifier_opacus_issue_shared.ipynb

note: my code isn't the prettiest, but should be pretty clear.  It was hacked together, just to test this issue. It was adapted from the image-classifier-tutorial : https://opacus.ai/tutorials/building_image_classifier .  
(I'd rather share ugly code so that you can look into the issue, sooner, but let me know if parts are unclear.)

RoyRin:
[...]

PR #338, opened 2022-02-01
Use correct data loader variable in cifar tutorial

ffuuugor:
As correctly pointed by @RoyRin in #308, image classifier tutorial doesn't use the data loader returned from the `make_private` (which it should)

### DIFF
--- a/opacus/validators/module_validator.py
+++ b/opacus/validators/module_validator.py
@@ -74,6 +74,7 @@ class ModuleValidator:
         """
         return len(cls.validate(module, strict=False)) == 0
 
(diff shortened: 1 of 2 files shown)


opacus/validators/module_validator.py
```diff
@@ -74,6 +74,7 @@ def is_valid(cls, module: nn.Module) -> bool:
         """
         return len(cls.validate(module, strict=False)) == 0
 
+    # TODO: fix method doesn't respect devices: new modules are always created on cpu
```
karthikprasad, line 77: Great catch! let me create an issue to track this

karthikprasad (approved):

PR #347, opened 2022-02-04
Fix per-layer clipping in distributed

alexandresablayrolles:
### Problem

`DDPPerLayerOptimizer` was buggy and our test did not catch that. 

### Changes

Per-layer clipping uses the regular pytorch DDP (Distributed Data Parallel) instead of our custom DPDDP (Differentially Private DDP). Our PrivacyEngine only checked for DPDDP and not for DDP.
I fixed PrivacyEngine to have distributed=True if wrapper is DDP instead of DPDDP. 

This made the test non-buggy and I modified DPDDP to pass the test (adding `_check_skip_next_step` on step, adding `if self._step_skip_queue and self._step_skip_queue[0]` when doing per-parameter noise)

### DIFF
--- a/opacus/optimizers/ddp_perlayeroptimizer.py
+++ b/opacus/optimizers/ddp_perlayeroptimizer.py
@@ -15,7 +15,7 @@
 from __future__ import annotations
 
 from functools import partial
-from typing import List, Optional
+from typing import Callable, List, Optional
 
 import torch
 from torch import nn
@@ -29,8 +29,7 @@ def _clip_and_accumulate_parameter(p: nn.Parameter, max_grad_norm: float):
     per_sample_clip_factor = (max_grad_norm / (per_sample_norms + 1e-6)).clamp(max=1.0)
 
     grad = torch.einsum("i,i...", per_sample_clip_factor, p.grad_sample)
-
-    if hasattr(p, "summed_grad"):
+    if p.summed_grad is not None:
         p.summed_grad += grad
     else:
         p.summed_grad = grad
@@ -47,14 +46,15 @@ class DistributedPerLayerOptimizer(DPOptimizer):
         optimizer: Optimizer,
         *,
         noise_multiplier: float,
-        max_grad_norms: List[float],
+        max_grad_norm: List[float],
         expected_batch_size: Optional[int],
         loss_reduction: str = "mean",
         generator=None,
         secure_mode: bool = False,
     ):
         self.rank = torch.distributed.get_rank()
-        self.max_grad_norms = max_grad_norms
+        self.world_size = torch.distributed.get_world_size()
+        self.max_grad_norms = max_grad_norm
         max_grad_norm = torch.norm(torch.Tensor(self.max_grad_norms), p=2).item()
         super().__init__(
             optimizer,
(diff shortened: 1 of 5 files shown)


opacus/privacy_engine.py
```diff
@@ -327,7 +328,7 @@ def make_private(
         if noise_generator and self.secure_mode:
             raise ValueError("Passing seed is prohibited in secure mode")
 
-        distributed = type(module) is DPDDP
+        distributed = type(module) is DPDDP or type(module) is DDP
```
romovpa, line 331: Actually, we should avoid type comparisons like this. Do this instead:

```
isinstance(module, (DPDDP, DDP))
```

opacus/tests/multigpu_gradcheck.py
```diff
@@ -53,20 +53,23 @@ def forward(self, x):
 def demo_basic(rank, weight, world_size, dp, clipping):
     torch.manual_seed(world_size)
     batch_size = 32
-    withdp = "with" + ("out " if not dp else "")
-    print(f"Running basic DDP {withdp} differential privacy example on rank {rank}.")
+    withdp = "with" + ("out" if not dp else "")
+    print(
+        f"Running {clipping}-clipping DDP {withdp} differential privacy example on rank {rank}."
+    )
```
romovpa, line 59: We shouldn't use prints anywhere in the library, in tests too. It may have unexpected side effects. 
If we really needs to do this output here, we should [use logging](https://docs.python.org/3/howto/logging.html):
```python
import logging
logging.info(f"Running {clipping}-clipping DDP {withdp} differential privacy example on rank {rank}.")
```

opacus/tests/multigpu_gradcheck.py
```diff
@@ -77,7 +80,7 @@ def demo_basic(rank, weight, world_size, dp, clipping):
     if dp:
         max_grad_norm = 1e8
```
romovpa, line 81: Let's follow [pep-8](https://www.python.org/dev/peps/pep-0008/#constants) and name this constant with capital letters: `MAX_GRAD_NORM`. This will emphasize that this thing is pre-defined and hard coded.

romovpa (changes requested):
Added some comments. 
Also, @alexandresablayrolles may I ask you to describe in a few words the problem and the proposed solution? I didn't work on DDP and it's hard for me to validate the logic.

opacus/tests/multigpu_gradcheck.py
alexandresablayrolles: It's not a constant: if clipping is per-layer, it becomes a list, see line 80.

opacus/optimizers/ddp_perlayeroptimizer.py
```diff
@@ -94,20 +100,30 @@ def clip_and_accumulate(self):
...
 
+        self._is_last_step_skipped = False
+        return True
+
     def _ddp_per_layer_hook(
         self, p: nn.Parameter, max_grad_norm: float, _: torch.Tensor
     ):
         _clip_and_accumulate_parameter(p, max_grad_norm)
+        # Equivalent ot _check_skip_next_step but without popping because it has to be done for every parameter p
+        if self._step_skip_queue and self._step_skip_queue[0]:
```
romovpa, line 124: This will crash if `self._step_skip_queue = []`

romovpa: `self._step_skip_queue is not None and len(self._step_skip_queue) > 0 and ...`

romovpa: Add argument to `self._check_skip_next_step(pop_next=False)`

romovpa (changes requested):

romovpa (approved):

Issue #346, opened 2022-02-02
The Sigma binary search is not conservative and can overshoot target_epsilon


tholop:
## 🐛 Small bug/unexpected behavior

There is a binary search to compute the value of `sigma` necessary to reach a given `target_epsilon` here: https://github.com/pytorch/opacus/blob/579546053050827fa9d71587c87ac477983db17b/opacus/accountants/utils.py#L55
The search stops when the search interval is smaller than `SIGMA_PRECISION`, which is 0.01 by default. 

The problem is that the value returned for sigma can be anywhere in `[sigma_opt - SIGMA_PRECISION/2, sigma_opt + SIGMA_PRECISION/2]`, where sigma_opt is the value that would give us exactly `target_epsilon`. As a result, the binary search sometimes overshoots `sigma_opt` and returns a value of `sigma` that yields `epsilon > target_epsilon`.

This is expected for a standard binary search, but for privacy budget I would expect a conservative binary search that returns a value in `[sigma_opt, sigma_opt + SIGMA_PRECISION]`, to make sure that sigma doesn't over-consume privacy. This is not very impactful for small values of `SIGMA_PRECISION` but the error is not completely negligible (my example below shows an over-consumption of 0.5 for epsilon).

If this is the intended behavior, I think that the docstring should make it clear. Right now, it claims that it returns "The noise level sigma to ensure privacy budget of (target_epsilon, target_delta)".

## To Reproduce

Copy the following lines at the end of `utils.py`:
```python
def main():
    
    sample_rate = 0.01
    epochs = 100
    target_delta = 1e-5
    target_epsilon = 5.0
    
    sigma = get_noise_multiplier(target_epsilon=target_epsilon, target_delta=target_delta, sample_rate=sample_rate,epochs=epochs, accountant="rdp")
    accountant = create_accountant(mechanism="rdp")
    accountant.steps = [(sigma, sample_rate, int(epochs / sample_rate))]
    
    actual_epsilon = accountant.get_epsilon(delta=target_delta)

    print(f"Target epsilon: {target_epsilon}.\n Sigma returned by binary search: {sigma}.\n Actual epsilon using this sigma: {actual_epsilon}")

if __name__ == "__main__":
   main()
```

Then, execute `python utils.py`:
```
Target epsilon: 5.0.
 Sigma returned by binary search: 1.1715283203125.
 Actual epsilon using this sigma: 5.055592282662654
```

## Expected behavior

[...]

PR #348, opened 2022-02-04
Conservative search for target epsilon in get_noise_multiplier

alexandresablayrolles:
## Problem

@tholop noticed in #346 that `get_noise_multiplier` can overshoot the target epsilon when searching for highest possible noise multiplier. This is because of the binary search implementation.

## Changes

- Noise multiplier now verifies that the actual epsilon is *lower* than the target epsilon.
  - Added test for this.  
- Bounds of the range of search for sigma, and precision can now be given as arguments to `get_noise_multiplier`.

Differential Revision: D34003401

### DIFF
--- a/opacus/accountants/utils.py
+++ b/opacus/accountants/utils.py
@@ -13,26 +13,19 @@
 # limitations under the License.
 
 from opacus.accountants import create_accountant
+from typing import Optional
 
-
-# min range bound when searching for sigma given epsilon
-DEFAULT_SIGMA_MIN_BOUND = 0.01
-# starting point for a max range bound when searching for sigma given epsilon
-DEFAULT_SIGMA_MAX_BOUND = 10
-# condition to halt binary search for sigma given epsilon
-SIGMA_PRECISION = 0.01
-# max possible value for returned sigma.
-# Noise higher than MAX_SIGMA considered unreasonable
-MAX_SIGMA = 2000
-
+MAX_SIGMA = 1e6
 
 def get_noise_multiplier(
     *,
     target_epsilon: float,
     target_delta: float,
     sample_rate: float,
-    epochs: int,
+    epochs: Optional[int] = None,
+    steps: Optional[int] = None,
     accountant: str = "rdp",
+    epsilon_tolerance: float = 0.01,
     **kwargs,
 ) -> float:
     r"""
@@ -44,7 +37,9 @@ def get_noise_multiplier(
         target_delta: the privacy budget's delta
         sample_rate: the sampling rate (usually batch_size / n_data)
         epochs: the number of epochs to run
+        steps: number of steps to run
         accountant: accounting mechanism used to estimate epsilon
+        epsilon_tolerance: precision for the binary search
     Returns:
         The noise level sigma to ensure privacy budget of (target_epsilon, target_delta)
     """
@@ -53,27 +48,33 @@ def get_noise_multiplier(
         raise NotImplementedError(
             "get_noise_multiplier is currently only supports RDP accountant"
         )
+    if (steps is None) == (epochs is None):
+        raise ValueError(
+            "get_noise_multiplier takes as input EITHER a number of steps or a number of epochs"
+        )
+    if steps is None:
+        steps = int(epochs / sample_rate)
 
-    eps = float("inf")
-    sigma_min = DEFAULT_SIGMA_MIN_BOUND
-    sigma_max = DEFAULT_SIGMA_MAX_BOUND
+    eps_high = float("inf")
     accountant = create_accountant(mechanism=accountant)
 
-    while eps > target_epsilon:
-        sigma_max = 2 * sigma_max
-        accountant.steps = [(sigma_max, sample_rate, int(epochs / sample_rate))]
-        eps = accountant.get_epsilon(delta=target_delta, **kwargs)
-        if sigma_max > MAX_SIGMA:
+    sigma_low, sigma_high = 0, 10
+    while eps_high > target_epsilon:
+        sigma_high = 2 * sigma_high
+        accountant.steps = [(sigma_high, sample_rate, int(epochs / sample_rate))]
+        eps_high = accountant.get_epsilon(delta=target_delta, **kwargs)
+        if sigma_high > MAX_SIGMA:
             raise ValueError("The privacy budget is too low.")
 
-    while sigma_max - sigma_min > SIGMA_PRECISION:
-        sigma = (sigma_min + sigma_max) / 2
-        accountant.steps = [(sigma, sample_rate, int(epochs / sample_rate))]
+    while target_epsilon - eps_high > epsilon_tolerance:
+        sigma = (sigma_low + sigma_high) / 2
+        accountant.steps = [(sigma, sample_rate, steps)]
         eps = accountant.get_epsilon(delta=target_delta, **kwargs)
 
         if eps < target_epsilon:
-            sigma_max = sigma
+            sigma_high = sigma
+            eps_high = eps
         else:
-            sigma_min = sigma
+            sigma_low = sigma
 
-    return sigma
+    return sigma_high
--- a/opacus/tests/accountants_test.py
+++ b/opacus/tests/accountants_test.py
@@ -15,7 +15,9 @@
 
 import unittest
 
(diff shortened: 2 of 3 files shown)


opacus/accountants/utils.py
```diff
@@ -39,27 +38,31 @@ def get_noise_multiplier(
...
 
     while eps > target_epsilon:
-        sigma_max = 2 * sigma_max
-        accountant.steps = [(sigma_max, sample_rate, int(epochs / sample_rate))]
+        sigma_high = 2 * sigma_high
+        accountant.steps = [(sigma_high, sample_rate, steps)]
         eps = accountant.get_epsilon(delta=target_delta, **kwargs)
-        if sigma_max > MAX_SIGMA:
+        if sigma_high > sigma_max:
             raise ValueError("The privacy budget is too low.")
```
romovpa, line 56: Here we try to automatically initialize `sigma_high` for the binary search. This behaviour is not consistent with the docstring: I can set a specific bound but it can change it anyway. Also, having both `sigma_high` and `sigma_max` is counter-intuitive and it takes knowing the whole algorithm to understand what is the difference.

## Proposal

Leave only `sigma_low`, `sigma_high` in parameters. I would also rename to min/max to emphasize that the result will be bounded:

```python
def get_noise_multiplier(
    ...
    sigma_min: float = 0.01,
    sigma_max: Optional[float] = None,
)
```

If `sigma_max is None`, run this part to get the upper bound starting from `sigma_max+1`. 

Leave MAX_SIGMA a global constant, maybe make it higher.

opacus/accountants/utils.py
```diff
@@ -1,24 +1,18 @@
...
     target_delta: float,
     sample_rate: float,
-    epochs: int,
+    epochs: int = None,
+    steps: int = None,
     accountant: str = "rdp",
+    sigma_precision=0.01,
+    sigma_low=0.01,
+    sigma_high=10,
+    sigma_max=2000,
```
romovpa, line 15: Add type definitions please

opacus/accountants/utils.py
```diff
@@ -1,24 +1,18 @@
...
 def get_noise_multiplier(
     *,
     target_epsilon: float,
     target_delta: float,
     sample_rate: float,
-    epochs: int,
+    epochs: int = None,
+    steps: int = None,
     accountant: str = "rdp",
+    sigma_precision=0.01,
```
romovpa, line 12: Let's rename it to `sigma_tolerance`, which is the common word used in optimization (see [scipy.bisect](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.bisect.html)).

Isn't this precision too low? I mean we should make this value lower, closer to typical float precision. Btw, this higher/lower confusion is another reason to rename the argument to tolerance.

opacus/accountants/utils.py
```diff
@@ -1,24 +1,18 @@
...
-
-
 def get_noise_multiplier(
     *,
     target_epsilon: float,
     target_delta: float,
     sample_rate: float,
-    epochs: int,
+    epochs: int = None,
+    steps: int = None,
```
romovpa, line 10: If the argument can be None, it should have type `Optional[int]`

opacus/accountants/utils.py
```diff
@@ -39,27 +38,31 @@ def get_noise_multiplier(
...
 
         if eps < target_epsilon:
-            sigma_max = sigma
+            sigma_high = sigma
         else:
-            sigma_min = sigma
+            sigma_low = sigma
 
-    return sigma
+    return sigma_high
```
romovpa, line 68: A possible corner case when `sigma_high < sigma_low` is not considered here.

opacus/accountants/utils.py
```diff
@@ -39,27 +38,31 @@ def get_noise_multiplier(
...
+        sigma = (sigma_low + sigma_high) / 2
+        accountant.steps = [(sigma, sample_rate, steps)]
         eps = accountant.get_epsilon(delta=target_delta, **kwargs)
 
         if eps < target_epsilon:
-            sigma_max = sigma
+            sigma_high = sigma
         else:
-            sigma_min = sigma
+            sigma_low = sigma
```
romovpa, line 66: I'm thinking if we can use [`scipy.optimzie.bisect()`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.bisect.html) instead of implementing our own.
It is easy to implement binary search incorrectly so we need to make tests for all corner cases. If we use scipy, we don't need to add our own tests, plus it is much more stable than the rest of our dependencies.

opacus/accountants/utils.py
romovpa, line 12: You need to add check that `sigma_tolerance` is strictly positive. Alternative: pass is to scipy.bisect and let it do the check for you.

romovpa (changes requested):

opacus/accountants/utils.py
```diff
@@ -53,27 +51,40 @@ def get_noise_multiplier(
...
+    if eps_high > target_epsilon:
+        raise ValueError(
+            f"Noise multiplier of {sigma_high} is not enough to have a privacy budget of {target_epsilon}, please use a higher sigma_high."
+        )
+
+    accountant.steps = [(sigma_low, sample_rate, steps)]
+    eps = accountant.get_epsilon(delta=target_delta, **kwargs)
+    if eps < target_epsilon:
+        warnings.warn(
+            f"Noise multiplier of {sigma_low} is still too high for a privacy budget of {target_epsilon}, please use a lower sigma_low."
```
romovpa, line 75: I would replace "please use a lower sigma_low" with something like:
"you can benefit quality by allowing smaller sigma_low without exceeding target epsilon"

But the current version is also fine.

opacus/accountants/utils.py
```diff
@@ -13,26 +13,20 @@
...
     *,
     target_epsilon: float,
     target_delta: float,
     sample_rate: float,
-    epochs: int,
+    epochs: Optional[int] = None,
+    steps: Optional[int] = None,
     accountant: str = "rdp",
+    epsilon_tolerance: float = 0.01,
+    sigma_low: float = 0.01,
```
romovpa, line 28: I wanted to ask, why the default `sigma_low` is so high. We will surprise the user with the warning and suboptimal result if the best sigma is 1e-3 for example.
But then I found that I don't understand what is the point of having this parameter at all? Why not starting with sigma_low=0 always? 

This won't hurt this expression `target_epsilon - eps_high > epsilon_tolerance`. We will never reach this value closer than tolerance/2.

opacus/accountants/utils.py
```diff
@@ -13,26 +13,20 @@
...
+    sigma_high: float = 200,
```
romovpa, line 29: Why we are so sure that 200 is high enough? 

Why not eliminate this constant by probing `sigma_high = tol*2^k` using this removed algorithm:
```
    sigma_high = tol
    while eps > target_epsilon:
        sigma_high = 2 * sigma_high
        accountant.steps = [(sigma_high, sample_rate, int(epochs / sample_rate))]
        eps = accountant.get_epsilon(delta=target_delta, **kwargs)
        if sigma_high > MAX_SIGMA:
            raise ValueError("The privacy budget is too low.")
```

romovpa (changes requested):

opacus/accountants/utils.py
```diff
@@ -13,26 +13,19 @@
...
-DEFAULT_SIGMA_MIN_BOUND = 0.01
-# starting point for a max range bound when searching for sigma given epsilon
-DEFAULT_SIGMA_MAX_BOUND = 10
-# condition to halt binary search for sigma given epsilon
-SIGMA_PRECISION = 0.01
-# max possible value for returned sigma.
-# Noise higher than MAX_SIGMA considered unreasonable
-MAX_SIGMA = 2000
-
+MAX_SIGMA = 1e6
```
romovpa, line 18: There was a comment before which I recommend to leave because the constant looks a bit surprising.

```python
# Noise level higher than MAX_SIGMA considered unreasonable 
# leading to "not sufficient privacy budget" exception
MAX_SIGMA = 1e6
```

romovpa (approved):
Sweet! 
One nit to add a comment.

Issue #340, opened 2022-02-01
Easy way to specify fixer for BatchNorm


karthikprasad:
## 🚀 Feature
Make it easy to choose the replacement module for `BatchNorm` out of the available options.

## Motivation

The `BatchNorm` module is not DP compatible and hence Opacus recommends replacing it with `GroupNorm`. Unfortunately, this recommendation does not work out of the box for all use cases (eg #308). In addition to `GroupNorm`, Opacus does offer `InstanceNorm` as a replacement via fixer override; however, this route is prohibitively cumbersome.

While it is simple and preferable to push one default recommendation and allow for other choices through decorator registration, this is hard to justify especially when we have evidence of cases where default recommendation is sub-optimal.

## Pitch
The easiest approach would be to include a `kwargs` in `fix()` method and supply a `replace_bn_with_in` or some such boolean flag.
Eg:

```python
@register_module_fixer(
    [nn.BatchNorm1d, nn.BatchNorm2d, nn.BatchNorm3d, nn.SyncBatchNorm]
)
def fix(module: BATCHNORM, **kwargs) -> Union[nn.GroupNorm, INSTANCENRM]:
    is_replace_bn_with_in = kwargs.get("replace_bn_with_in", False)
    return (_batchnorm_to_groupnorm(module) if not is_replace_bn_with_in else _batchnorm_to_instancenorm(module))
```

Though simple, this introduces yet another flag that users will have to remember. There is also no easy way of documenting this flag in the API without also exposing this BatchNorm fixer specific flag to `ModuleValidator.fix()`.

## Alternatives

Unable to think of any simpler approach. Please comment if you have any ideas.

PR #353, opened 2022-02-09
Add replace_bn_with_in boolean flag to BatchNorm fix method

karthikprasad:
Summary: add `replace_bn_with_in` boolean flag to `fix()`and select InstanceNorm accordingly

Differential Revision: D34088182


karthikprasad:
closes #340

PR #356, opened 2022-02-10
clone replacement modules on the same device as original

### LINKED ISSUE #339: fix method doesn't respect devices: new modules are always created on cpu
_Originally posted by @karthikprasad in https://github.com/pytorch/opacus/pull/338#discussion_r796273771_

karthikprasad:
Summary: closes #339

Differential Revision: D34132507

### DIFF
--- a/opacus/validators/module_validator.py
+++ b/opacus/validators/module_validator.py
@@ -86,7 +86,6 @@ class ModuleValidator:
         """
         return len(cls.validate(module, strict=False)) == 0
 
-    # TODO: fix method doesn't respect devices: new modules are always created on cpu
     @classmethod
     def fix(cls, module: nn.Module, **kwargs) -> nn.Module:
         """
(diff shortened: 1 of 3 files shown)


opacus/validators/module_validator.py
```diff
@@ -112,6 +111,8 @@ def fix(cls, module: nn.Module, **kwargs) -> nn.Module:
                 # get a repalcement for sub_module
                 sub_module_fixer = ModuleValidator.FIXERS[type(sub_module)]
                 new_sub_module = sub_module_fixer(sub_module, **kwargs)
+                # move new_sub_module to the same device as that of sub_module
```
ffuuugor, line 114: Shouldn't we handle empty parameter list here as well?

ffuuugor (changes requested):
If I'm reading this correctly, `test_fix` should fail when executed on a GPU machine (but I don't think we run unit test on GPUs in CircleCI)

opacus/validators/module_validator.py
ffuuugor: Yup, you're right

ffuuugor (approved):
Should've approved it the first time, my bad

Issue #352, opened 2022-02-08
In opacus.data_loader, batch_sampler option is mutually exclusive with drop_last


Varal7:
## 🐛 Bug

In opacus.data_loader, batch_sampler option is mutually exclusive with drop_last

## To Reproduce

```python
import torch
from torch.utils.data import DataLoader, TensorDataset
from opacus.data_loader import DPDataLoader
x, y = torch.randn(63, 5), torch.randint(0, 2, (63,))
dataset = TensorDataset(x,y)
data_loader = DataLoader(dataset, batch_size=4, drop_last=True)
dp_data_loader = DPDataLoader.from_data_loader(data_loader)
```

```
ValueError: batch_sampler option is mutually exclusive with batch_size, shuffle, sampler, and drop_last
```

## Expected behavior

Opacus should issue a warning that `drop_last` is not compatible with `DPDataLoader` and not propagate it to `DataLoader.__init__()`.

romovpa:
@Varal7 thanks for reporting!
Fair enough. I agree we should ignore `drop_last` when converting DataLoader. 
I'm not sure if it's necessary to issue a warning though. The transformation is explicit, I think it's enough to make sure it is clear from the docstring.

romovpa:
Fixed in #357

PR #357, opened 2022-02-14
Warn and ignore "drop_last" when set in DPDataLoader

gautham-kollu:
Summary: Add a warning to show "drop_last" is being ignored and is not compatible with DPDataLoader

Differential Revision: D34197820

### DIFF
--- a/opacus/data_loader.py
+++ b/opacus/data_loader.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 from typing import Any, Optional, Sequence
 
 import torch
@@ -23,6 +24,7 @@ from torch.utils.data import BatchSampler, DataLoader, Dataset, IterableDataset,
 from torch.utils.data._utils.collate import default_collate
 from torch.utils.data.dataloader import _collate_fn_t, _worker_init_fn_t
 
+logger = logging.getLogger(__name__)
 
 def wrap_collate_with_empty(
     collate_fn: Optional[_collate_fn_t], sample_empty_shapes: Sequence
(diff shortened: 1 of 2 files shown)


opacus/data_loader.py
```diff
@@ -143,13 +144,15 @@ def __init__(
         if collate_fn is None:
             collate_fn = default_collate
 
+        if drop_last:
+            warnings.warn("Ignoring drop_last as it is not compatible with DPDataLoader.")
```
karthikprasad, line 148: nit: As per @romovpa's comment on #352, warning is probably not necessary as the client doesn't have to take any action. How about changing this to logger.warning() instead?

opacus/tests/dpdataloader_test.py
```diff
@@ -45,3 +45,10 @@ def test_collate_tensor(self):
         (s,) = next(iter(data_loader))
 
         self.assertEqual(s.size(0), 0)
+
+    def test_drop_last_true(self):
+        x = torch.randn(self.data_size, self.dimension)
+
+        dataset = TensorDataset(x)
+        data_loader = DataLoader(dataset, drop_last=True)
+        dp_data_loader = DPDataLoader.from_data_loader(data_loader)
```
karthikprasad, line 54: nit: pacify linter
```suggestion
        _ = DPDataLoader.from_data_loader(data_loader)
```

karthikprasad (commented):
Hi @gautham-kollu , thanks for taking this up. I just added a couple of nits, looks good otherwise. Happy to approve after the changes :)

karthikprasad (approved):
Thanks again!

PR #365, opened 2022-03-11
Refactor DPMultiheadAttention to match that of nn.MultiheadAttention.

paksas:
Summary:
After trying out several refactoring strategies, including the use of ParamRenameModule, I settled for replacing the linear layers with tracked tensors that closely model the state of nn.MultiheadAttention.

The main driver behind this decision was the duality of representations used by nn.MultiheadedAttention to represent the state of the packed and the unpacked input projection layers.

In order to use this new state representation I refactored the calculations of the input projections to use the same helper functions that nn.MultiheadAttention uses - torch.nn.functional._in_projection_packed and torch.nn.functional._in_projection.

The same refactoring strategy was used to replace the bias tensors applied to the transformed keys and queries (formerly represented by the seq_bias_k and seq_bias_v parameters). The main obstacle there was the shape mismatch of the representations.
In the process of refactoring, SequenceBias became obsolete and was removed.

DPMultiheadAttention_test.test_attn was enabled.

Differential Revision: D34719599


ffuuugor:
Following the internal discussion, closing the diff.
It has some good ideas, but by itself won't lead to a workable solution

Issue #123, opened 2021-01-22
DPMultiheadAttention is not drop-in compatible with nn.MultiheadAttention


Darktex:
## 🐛 Bug
DPMultiheadAttention is not drop-in compatible with nn.MultiheadAttention.

Opacus can only support certain "building block" layers (such as nn.Linear and nn.Conv2d), and layers written using those building blocks - in theory pretty much all of them, but in practice this leads to reimplementations. For example, while you can indeed write a LSTM using nn.Linear layers, the implementation we have in nn.LSTM doesn't do that (so that it can fuse ops and be faster). Whenever this happens, we reimplement the layer and make the update available in the Layers folder.

The aim of these reimplementations is that they all should be perfect drop-in replacements of their "normal PyTorch" counterparts, meaning that they should have exactly the same API, and exactly the same state_dict, so that users can train using the Opacus version, but then they can export the state_dict and load it for inference in a PyTorch environment that doesn't even have Opacus installed.

## Please reproduce using our [template Colab](https://colab.research.google.com/drive/1pCyN3_gN3WyhyQt8--eBPN5zVf2XZxiN?usp=sharing) and post here the link

Colab to repro this issue: https://colab.research.google.com/drive/1N4lqMmG21cROzx6qgJphTniMzcaX190C?usp=sharing

## To Reproduce

1. Instantiate a DPMultiheadAttention module
2. Instantiate a nn.MultiheadAttention with the same parameters
3. Their state_dicts should match!

## Additional context
There can be two reasons why this happens:

1. You have exactly the same matrices, but the names of the actual parameters are different due to how you organized your class.
2. You structured your layers differently so eg you have more/fewer matrices than the reference implementation

An example of 1 is the [DPLSTM](https://github.com/pytorch/opacus/blob/master/opacus/layers/dp_lstm.py). It's a big file, but essentially it has the same matrices as `nn.LSTM`, but they would be called something like `layers[0].ih` instead of `ih_0`. To fix this, we have an helper class named [ParamRenameModule](https://github.com/pytorch/opacus/blob/master/opacus/layers/param_rename.py) that allows you to specify a mapping. It's being used in DPLSTM, and it can help you here too.

In addition to this, here we also have Reason 2: we have a different number of matrices so we should align them. There are a few tricks that can help you here (this will teach you how `torch.nn.Module` works too!). Take a look at this code snippet to learn how PyTorch finds and registers parameters.

```python
class MyCoolModule(nn.Module):
    def __init__(self):
[...]

PR #387, opened 2022-03-15
Define a module for packed input projection

paksas:
Summary:
Added a new module that projects the three inputs of a multihead attention module. In contrast to the unpacked version, this module uses an operator that works with only two tensors.

The operator employed is the same one that is used by nn.MultiheadAttention in the case when the shapes of all three inputs match.

Differential Revision: D34850577


ffuuugor:
Closing, as this PR was created for demo purposes with no plans to merge (it's linked in #123 )

PR #407, opened 2022-04-05
Enable integration tests with the latest PyTorch by installing csprng from the source

romovpa:
## Problem

Integration tests depends on [`csprng`](https://github.com/pytorch/csprng). The latest csprng release was made in Mar 2021 and has strict dependency on `pytorch==1.8.0` while the code actually works fine with the newer versions of pytorch.

The release cycle is currently broken and there is no easy way to make a new release with the fix.

## Temporary solution

Instead of installing `csprng` from PyPI (having outdated broken release), for the purpose of integration tests we install it from the github source. Recommend doing the same in FAQ. 

This removes artificially created dependency on the earlier pytorch version and unblocks new features.

## Further work

As agreed with the former csprng developer, the Opacus team is the new owner of [`csprng`](https://github.com/pytorch/csprng) repository. We are going to update and fix the release cycle. After that csprng can be pip installed without wrong dependencies.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,7 +46,7 @@ commands:
       - run:
           name: "Install PyTorch"
           command: |
-            pip install torch torchvision torchaudio --extra-index-url https://download.pytorch.org/whl/cu114
+            pip install torch torchvision torchaudio 
 
   lint_flake8:
     description: "Lint with flake8"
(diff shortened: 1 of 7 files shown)


ffuuugor:
Do we plan to land this or can we close this one and continue in #419 ?

romovpa:
Closing this until csprng is fixed

Issue #182, opened 2021-06-15
Dilation parameter in Conv3d


ChrisWaites:
## 🐛 Bug

Currently, the dilation parameter does not work for the Conv3d layer, and it would be nice to be able to replicate existing video models with minimal changes. Not currently covered by the [tests](https://github.com/pytorch/opacus/blob/244265582bffbda956511871a907e5de2c523d86/opacus/tests/grad_samples/conv3d_test.py#L25).

## Please reproduce using our [template Colab](https://colab.research.google.com/drive/1pCyN3_gN3WyhyQt8--eBPN5zVf2XZxiN?usp=sharing) and post here the link

https://colab.research.google.com/drive/1GO_5prcipQPyIdngRENYaBgd-z2kGlZ9?usp=sharing

Current error:

```
...
/usr/local/lib/python3.7/dist-packages/opacus/utils/tensor_utils.py in unfold3d(tensor, kernel_size, padding, stride, dilation)
    121 
    122     if dilation != (1, 1, 1):
--> 123         raise NotImplementedError(f"dilation={dilation} not supported. We'd love a PR!")
    124 
    125     batch_size, channels, _, _, _ = tensor.shape

NotImplementedError: dilation=[2, 2, 2] not supported. We'd love a PR!
```

## Expected behavior

The given Colab should run without failure. Things fail with `dilation=[1, 1, 1]`, (which is a bug; this is equivalent to `dilation=1`, which is already supported). Also fails with `dilation=[2, 2, 2]` (more involved change).

## Additional context

A nice visualization of how dilation should work is given [here](https://github.com/vdumoulin/conv_arithmetic/blob/master/README.md#dilated-convolution-animations).

PR #408, opened 2022-04-06
Implementing 3D dilation

shatayu:
Summary:
Implemented 3D dilation for values that aren't `(1, 1, 1)` as requested by [this issue](https://github.com/pytorch/opacus/issues/182).

## The Algorithm
Algorithm works by dilating the kernel (see below) and then removing extraneous points after the tensor has been unfolded with the dilated kernel.  Removing extraneous points is done by calculating which rows/columns/etc. would have had zeros on them with the dilated kernel and removing them.

## Dilation
The dilated kernel size (for one dimension) is `kernel_size + (kernel_size - 1) * (dilation - 1)`. This is best explained through visualization (in 1D). Suppose we have a kernel that is 3 units long and we wish to dilate it by 3. Let `x` represent the units originally in the kernel and `o` represent the units introduced via dilation.

```
xooxoox
```

There are `kernel_size - 1` places to insert units from dilation and we want to insert `dilation - 1` units in each place. This leads to `(kernel_size - 1) * (dilation - 1)` total units from dilation and `kernel_size` original units, making a total `kernel_size + (kernel_size - 1) * (dilation - 1)` units.

Differential Revision: D35381703

### DIFF
--- a/opacus/utils/tensor_utils.py
+++ b/opacus/utils/tensor_utils.py
@@ -195,9 +195,6 @@ def unfold3d(
     if isinstance(dilation, int):
         dilation = (dilation, dilation, dilation)
 
-    if dilation != (1, 1, 1):
-        raise NotImplementedError(f"dilation={dilation} not supported. We'd love a PR!")
-
     batch_size, channels, _, _, _ = tensor.shape
 
     # Input shape: (B, C, D, H, W)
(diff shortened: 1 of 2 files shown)


opacus/utils/tensor_utils.py
```diff
@@ -221,3 +228,53 @@ def unfold3d(
...
+    indices_to_keep = [
+        list(range(0, dilated_kernel_size[i], dilation[i])) for i in range(kernel_rank)
+    ]
+
+    tensor_np = tensor.numpy()
+
+    axis_offset = len(tensor.shape) - kernel_rank
+
+    for dim in range(kernel_rank):
+        tensor_np = np.take(
```
karthikprasad, line 276: Please make linter happy on this line. :)

Issue #413, opened 2022-04-12
Tied parameters of a model cause error with grad_sampler


huseyinatahaninan:
## 🐛 Bug

When a model has tied parameters (such as tied embedding layer of GPT-2 model), the current version of Opacus stacks the grad_samples in a list instead of summing them up the gradients coming from shared parameters:
https://github.com/pytorch/opacus/blob/3ca9a67a1784cdb053623f95df8e835cdf3b6d08/opacus/grad_sample/grad_sample_module.py#L53

This later causes an error at: `per_sample_norms = torch.stack(per_param_norms, dim=1).norm(2, dim=1)`
because here all parameters have grad_samples as tensors whereas the tied embedding has it as a list of tensors instead.

In the earlier version of Opacus (v0.15) there was this function `create_or_accumulate_grad_sample` that enabled us to register tied embedding layer for instance as it adds up the gradients:
https://github.com/pytorch/opacus/blob/13cf52de6e967ef9cb1167ad4d4981ee15b45697/opacus/grad_sample/utils.py#L75

When I change the lines 50-53 below in the new version and simply replace them with `p.grad_sample += p._current_grad_sample` it works fine because it adds up the gradients coming from shared parameters.
https://github.com/pytorch/opacus/blob/3ca9a67a1784cdb053623f95df8e835cdf3b6d08/opacus/grad_sample/grad_sample_module.py#L53

I am curious why this change to stacking to list was made and whether I am missing anything regarding the tied parameters case. Is there a better fix I can make without changing the internal code? Any help would be greatly appreciated!

## Please reproduce using our [template Colab](https://colab.research.google.com/drive/1R-dnKipK9LOVV4_oKbuHoq4VvGKGbDnd?usp=sharing) and post here the link
https://colab.research.google.com/drive/1OfD4Ubb010Fe0M4rg13LUytNKPpk_asu?usp=sharing

## To Reproduce

Simply following the colab should show the error:

1. Take GPT-2 model from Huggingface
2. Add DP components
3. Run an arbitrary sample forward-backward and call optimizer
4. Tied parameters have list of grad_samples instead of a single tensor accumulating gradients and this causes the error

## Expected behavior

Previously we were able to register such layers by using `create_or_accumulate_grad_sample` as in v0.15 https://github.com/pytorch/opacus/blob/13cf52de6e967ef9cb1167ad4d4981ee15b45697/opacus/grad_sample/utils.py#L75 which added the grad_samples but now I am unable to find a similar solution.

[...]

PR #417, opened 2022-04-21
Support tied parameters

ffuuugor:
## Solution

As reported in #413, some models (including huggingface GPT-2) use tied parameters - i.e. sharing the same parameter between modules by simply using the same Parameter object. 
Opacus gets confused - it treats the second set of grad_samples as a new batch and accumulates them by stacking, which is obviously wrong. Conceptually this is not very different from RNNs, and the gradients should be simply added up.

Natural way to support this is to tweak the condition on which we distinguish between recurrent calls and new batch. Currently we look at module-level activations - if we've run out of the activations, it means that the recurrent calls are over and the next grad sample will be from a new batch. This obviously doesn't work for shared parameters - as the condition is module-level, but the parameter belong to multiple modules.

What we can do instead is to have parameter-level counters (`p._forward_counter`). This counter would increase on forward pass and decrease on backward. When it gets to zero, in would indicate that the latest backward pass is fully unrolled and we should expect new forward pass with the new data. If I understand correctly, we do not and should not allow multiple forwards or backwards in a row.

## Tests

I've added a new unit test for this and also checked Colab from the bug report - all seem to be working.

### DIFF
--- a/opacus/grad_sample/grad_sample_module.py
+++ b/opacus/grad_sample/grad_sample_module.py
@@ -27,9 +27,11 @@ from opacus.utils.module_utils import requires_grad, trainable_modules
 
 logger = logging.getLogger(__name__)
 
+OPACUS_PARAM_MONKEYPATCH_ATTRS = ["_forward_counter", "_current_grad_sample"]
+
 
 def create_or_accumulate_grad_sample(
-    *, param: torch.Tensor, grad_sample: torch.Tensor, layer: nn.Module
+    *, param: torch.Tensor, grad_sample: torch.Tensor, max_batch_len: int
 ) -> None:
     """
     Creates a ``_current_grad_sample`` attribute in the given parameter, or adds to it
@@ -46,9 +48,6 @@ def create_or_accumulate_grad_sample(
     if hasattr(param, "_current_grad_sample"):
         param._current_grad_sample[: grad_sample.shape[0]] += grad_sample
     else:
-        # TODO: maybe set max_batch_len on a parameter level, so
-        # you don't need to pass layer here?
-        max_batch_len = layer.max_batch_len
         param._current_grad_sample = torch.zeros(
             torch.Size([max_batch_len]) + grad_sample.shape[1:],
             device=grad_sample.device,
@@ -125,6 +124,7 @@ class GradSampleModule(nn.Module):
 
         for p in self.parameters():
             p.grad_sample = None
+            p._forward_counter = 0
 
     def __getattr__(self, item):
         try:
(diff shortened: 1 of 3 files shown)


opacus/grad_sample/grad_sample_module.py
```diff
@@ -296,6 +303,9 @@ def capture_activations_hook(
             module.activations = []
         module.activations.append(forward_input[0].detach())  # pyre-ignore
 
+        for p in module.parameters():
```
karthikprasad, line 306: Do you think it will help to add a comment above this line mentioning this is for tied parameters? A link to the PR itself is even better. Your PR summary is stellar!

opacus/tests/privacy_engine_test.py
```diff
@@ -262,7 +262,9 @@ def test_compare_to_vanilla(self):
         for do_noise in (True, False):
             for do_clip in (True, False):
                 for use_closure in (True, False):
-                    with self.subTest(do_noise=do_noise, do_clip=do_clip):
+                    with self.subTest(
+                        do_noise=do_noise, do_clip=do_clip, use_closure=use_closure
```
karthikprasad, line 266: Great catch!

opacus/grad_sample/grad_sample_module.py
```diff
@@ -341,15 +351,19 @@ def capture_backprops_hook(
...
         for param, gs in grad_samples.items():
-            create_or_accumulate_grad_sample(param=param, grad_sample=gs, layer=module)
+            create_or_accumulate_grad_sample(
+                param=param, grad_sample=gs, max_batch_len=module.max_batch_len
+            )
+
+        for p in module.parameters():
+            p._forward_counter -= 1
+            if p._forward_counter == 0:
+                promote_current_grad_sample(p)
```
karthikprasad, line 361: nit: I suppose this can still be inside the `if len(module.activations) == 0` block?

karthikprasad (approved):
A1 code changes! ( a couple of nits)
A0 summary! <3

Thanks!

opacus/grad_sample/grad_sample_module.py
```diff
@@ -125,6 +124,7 @@ def __init__(
 
         for p in self.parameters():
```
karthikprasad, line 125: <Comment unrelated to this specific PR/issue.>
This had not caught my eye before. Is there any reason why we attach `.grad_sample` to all parameters instead of just the trainable ones?

opacus/grad_sample/grad_sample_module.py
ffuuugor: The check itself - yes, but the decrement - no.

opacus/grad_sample/grad_sample_module.py
ffuuugor: Weeell, I don't think there's a reason. 
I always thought we've had a util method that does '[p for p in model.parameters() if p.requires_grad]', but apparently no, and we're consistently using just `model.parameters()` all over

vdasu:
Is this functionality added to the latest version of Opacus? I noticed the PR is closed. I am facing the same issue when trying to train GPT-2

Issue #420, opened 2022-04-26
Bug in DPOptimizer: actual sample size is not equal to batch size


charlotte12l:
## To Reproduce
> :warning: **We cannot help you without you sharing reproducible code**. Do not ignore this part :)

1. add `print('actual batch:', len(per_sample_clip_factor)` in  `clip_and_accumulate`
2. add `print('add noise!')` in `add_noise`
3. add `print('scale!', 'expected_batch_size:', self.expected_batch_size , 'iter:', self.accumulated_iterations)` in `scale_grad`

## Expected behavior

`actual sample size` is not equal to `expected_batch_size`! However, in `scale_grad`, it is using the `expected_batch_size` for average. Shouldn't we use the `actual sample size` for averaging? 

```
actual sample size: 14
add noise!
scale! expected_batch_size: 16 iter: 1
actual sample size: 10
add noise!
scale! expected_batch_size: 16 iter: 1
actual sample size: 17
add noise!
scale! expected_batch_size: 16 iter: 1


```

Besides, things would get worse when we use `BatchMemoryManager`, because we'll often do **two** rounds of `clip_and_accumulate` then do one round of `add_noise` and `scale_grad`. The `expected_batch_size` for average is totally different from the sum of the sample size feeded into the `clip_and_accumulate`
```
    with BatchMemoryManager(
        data_loader=train_dataloader, 
        max_physical_batch_size=MAX_PHYSICAL_BATCH_SIZE, 
        optimizer=optimizer
    ) as memory_safe_data_loader:
        for step, batch in enumerate(tqdm(memory_safe_data_loader)):

```

## Environment

I reproduced this use your template colab: [link](https://colab.research.google.com/drive/1LTAC_X8OMQh-x3gdt5H0HI8l_5uaMEDm?usp=sharing)

karthikprasad:
Hi! The difference in actual batch size is expected and is due to Poisson sampling in DP Dataloader. See https://github.com/pytorch/opacus/blob/main/opacus/data_loader.py#L79-L82 
> ... it switches ``batch_sampler`` to ``UniformWithReplacementSampler``, thus enabling
    Poisson sampling (i.e. each element in the dataset is selected to be in the
    next batch with a certain probability defined by ``sample_rate`` parameter).
    NB: this typically leads to a batches of variable size.

charlotte12l:
[...]

PR #421, opened 2022-04-27
Implementation of adaptive clipping optimizer

charlotte12l:
## Types of changes

- What problem does it solve? 
  This is an implementation of the popular adaptive clipping [paper](https://arxiv.org/pdf/1905.03871.pdf), which is proven to increase performance while maintaining the same privacy budget. Google has implemented a federated version of this, and I think it would also be helpful to have a central version. Besides, the adaptive clipping methods were mentioned in previous issues as a potential solution: https://github.com/pytorch/opacus/issues/152, https://discuss.pytorch.org/t/question-regarding-gradient-clipping/114707

This is the [notebook](https://colab.research.google.com/drive/19VuiZ_Tb0PVlQ7RNhFEOKpom-k83bk0d?usp=sharing) that I tested on. Except for the new optimizer, other steps are using the Opacus tutorial.

Also, the reason why I kept asking the `actual sample size` stuff in https://github.com/pytorch/opacus/issues/420 (thank you for your detailed explanation!) is that in adaptive clipping, (based on my understanding) the quantile should be computed using the `actual sample size` , so here I added support for that. Feel free to correct me if I'm wrong!

## Checklist

- [x]  I have read the **CONTRIBUTING** document and completed the CLA (see **CONTRIBUTING**).
- [ ] All tests passed, and additional code has been covered with new tests.

cc @wanrongz


opacus/optimizers/adaclipoptimizer.py
```diff
@@ -0,0 +1,168 @@
...
+                "non-trivial gradient accumulation behaviour"
+            )
+
+        for p in self.params:
+            p.grad_sample = None
+
+            if not self._is_last_step_skipped:
+                p.summed_grad = None
+
+        self.original_optimizer.zero_grad(set_to_none)
```
alexandresablayrolles, line 97: nit: this can be simplified as super().zero_grad(set_to_none)

opacus/optimizers/adaclipoptimizer.py
```diff
@@ -0,0 +1,168 @@
...
+        ]
+        per_sample_norms = torch.stack(per_param_norms, dim=1).norm(2, dim=1)
+        per_sample_clip_factor = (self.max_grad_norm / (per_sample_norms + 1e-6)).clamp(
+            max=1.0
+        )
+
+        self.sample_size += len(per_sample_clip_factor)
+        self.unclipped_num += (
+            len(per_sample_clip_factor) - (per_sample_clip_factor < 1).sum()
+        )
```
alexandresablayrolles, line 114: nit: maybe we can add a comment before L111 saying that the two lines below are the only change relative to the parent DPOptimizer class.

opacus/optimizers/adaclipoptimizer.py
```diff
@@ -0,0 +1,168 @@
...
+            )
+            p.grad = (p.summed_grad + noise).view_as(p.grad)
+
+            _mark_as_processed(p.summed_grad)
+
+        noise = _generate_noise(
+            std=self.fraction_std,
+            reference=self.unclipped_num,
+            generator=self.generator,
+        )
```
alexandresablayrolles, line 147: Same here, I would call super().add_noise() as it improves readability

opacus/optimizers/adaclipoptimizer.py
```diff
@@ -0,0 +1,168 @@
...
+        """
+        unclipped_frac = self.unclipped_num / self.sample_size
+        self.max_grad_norm *= torch.exp(
+            -self.clipbound_learning_rate
+            * (unclipped_frac - self.target_unclipped_quantile)
+        )
+        if self.max_grad_norm > self.max_clipbound:
+            self.max_grad_norm = self.max_clipbound
+        elif self.max_grad_norm < self.min_clipbound:
+            self.max_grad_norm = self.min_clipbound
```
alexandresablayrolles, line 164: I think we can put all the code above in a function update_max_grad_norm(). While I think calling it under scale_grad works, it is not ideal as it is not the purpose of the scale_grad function.
I would recommend to add it in the pre_step, something like that:

```
def pre_step(
        self, closure: Optional[Callable[[], float]] = None
    ) -> Optional[float]:
    pre_step_full = super().pre_step()
    if pre_step_full:
        self.update_max_grad_norm()
    return pre_step_full
    
```

alexandresablayrolles (changes requested):
Thanks a lot for the PR @charlotte12l ! I have left a couple of comments that I think will increase readability, let me know what you think.

opacus/optimizers/adaclipoptimizer.py
charlotte12l: Thank you so much for your suggestions! All resolved!

alexandresablayrolles (approved):
Thanks for the changes, looks great!

romovpa:
Thanks @charlotte12l for contributing!
Is it the failed integration test that prevents us from merging this, or there is something else to be done? @alexandresablayrolles

charlotte12l:
@romovpa Thanks for commenting! I think the failed test is unrelated to this PR? I haven't add test for this new adaptive optimizer, and the failed test shown in commit [14380bc](https://email.circleci.com/c/eJwVj71uxCAQhJ_G7rD4ZylcpImUpI-uxcABFxwcjC1fnj5EmuJbjTSz42aQGsY0v98-jv3t8yuu4TothRv8hIlIwUFiTDjTZOB4K3szebKp2uxR9sHYJ_ouLd2Tr-gkCMTCnRYSAF2_Z3uMcdbeaNozGAWCNcbaUgLggGOpnFRqzHNsbdsH9jLQ166QWjyWyZa1H9uzlWpjp7IZe-wdurGm1qF_BXixjHN3px4LLohbqOYAgirHQHqivPFmrLONpubSms_pIPTqU8JqUv4v-QOot08g) is `opacus/tests/privacy_engine_test.py::PrivacyEngineTextTest::test_noise_level`. 

Now seems this PR was already merged to main(https://github.com/pytorch/opacus/commit/c9abdcf1a6180965b427421eb019550da76a03ff), and all the check has passed?

lucacorbucci:
Hi @charlotte12l I have a few questions regarding this PR. I'd like to use AdaClipDPOptimizer but I'm not sure about the parameters that I have to pass.

After reading the paper I think that target_unclipped_quantile should be the target quantile of the norm distribution at which we want to clip. This should be in the range [0,1].

I'm not sure about the other 4 parameters: 'clipbound_learning_rate', 'max_clipbound', 'min_clipbound', and 'unclipped_num_std'. 

I've seen in the notebook that you linked that you used the following parameters but I can't figure out how you choose them.

clipbound_learning_rate=0.1,
max_clipbound=10,
min_clipbound=0.05,
unclipped_num_std=BATCH_SIZE/20,

PR #423, opened 2022-05-03
zero_grad() between forward and backward passes generates error message

pierrestock:
Summary:
Fixing the previous fix in D36067469 (https://github.com/pytorch/opacus/commit/56df34ec10b380844b6f7f73267246a8e100d301).

Context regarding the execution order of backward hooks:
- `register_backward_hook` has **constant** behaviors across tested PyTorch versions
- `register_full_backward_hook` has various behaviors.

Previous fix was working for `torch==1.11` but not for `torch==1.8.1`, hence CircleCI failures. This fix now works for **both** versions.

Note that `register_backward_hook` will be deprecated in the future by the Pytorch team.

Differential Revision: D36105357

### DIFF
--- a/opacus/privacy_engine.py
+++ b/opacus/privacy_engine.py
@@ -28,17 +28,20 @@ from torch.nn.parallel import DistributedDataParallel as DDP
 from torch.utils.data import DataLoader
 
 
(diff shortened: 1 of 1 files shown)


romovpa (approved):
Thanks for fixing!

PR #426, opened 2022-05-05
fix opacus nn.functional import (#901)

ngimel:
Summary:
X-link: https://github.com/pytorch/benchmark/pull/901
Additional context: this change is necessitated by https://github.com/pytorch/pytorch/pull/73434, that replaces `import torch.nn.functional as F` with `import torch.nn.functional`, causing `from torch.functional import F` in opacus to no longer be valid.

per title

Differential Revision: D36172316

### DIFF
--- a/examples/imdb.py
+++ b/examples/imdb.py
@@ -22,10 +22,10 @@ import argparse
 import numpy as np
 import torch
 import torch.nn as nn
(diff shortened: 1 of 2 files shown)


karthikprasad (approved):
Thanks.

malfet (approved):

Issue #373, opened 2022-03-11
Support checkpoints


ffuuugor:
Current support for model saving/loading is somewhat clunky. The easiest way is to simply convert model to non-private version before checkpointing, and save privacy accounting info manually

This can be easily supported out of the box

See this [post](https://discuss.pytorch.org/t/how-to-store-the-state-and-resume-the-state-of-the-privacyengine/138538/3) for more context

PR #429, opened 2022-05-17
Support checkpoints

karthikprasad:
## Types of changes

#373 

Unit tests.

## Checklist

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -88,7 +88,7 @@ This statement extends to all downstream uses of this model: its inferences, fin
 
 From the expression above it is obvious that epsilon and delta play different roles: epsilon controls the multiplicative increase in the baseline probability while delta lifts all probabilities by the same amount. For instance, if your baseline scenario (the model trained on *D*′, without your data) assigns 0 probability to some event, the bound on observing this event on *D* (that includes your data) is delta. Because of that, we’d like to target epsilon to be a small constant and select delta to be tiny. A rule of thumb is to set delta to be less than the inverse of the size of the training dataset.
 
-Epsilon and delta are computed *ex post*, following an optimizer run. In fact, for each delta there’s some epsilon, depending on that delta, such that the run satisfies (epsilon, delta)-DP. The call `privacy_engine.get_privacy_spent(delta)` outputs that epsilon in its first return value.
+Epsilon and delta are computed *ex post*, following an optimizer run. In fact, for each delta there’s some epsilon, depending on that delta, such that the run satisfies (epsilon, delta)-DP. The call `privacy_engine.accountant.get_privacy_spent(delta=delta)` outputs that epsilon in its first return value.
 
 Importantly, (epsilon, delta)-DP is a *conservative upper bound* on the actual privacy loss. There’s [growing](https://arxiv.org/abs/2006.07709) [evidence](https://arxiv.org/pdf/2006.11601.pdf) that the observable privacy loss of the DP-SGD algorithm can be significantly smaller.
 
@@ -108,9 +108,9 @@ Opacus computes and stores *per-sample* gradients under the hood. What this mean
 
 Although we report expended privacy budget using the (epsilon, delta) language, internally, we track it using Rényi Differential Privacy (RDP) [[Mironov 2017](https://arxiv.org/abs/1702.07476), [Mironov et al. 2019](https://arxiv.org/abs/1908.10530)]. In short, (alpha, epsilon)-RDP bounds the [Rényi divergence](https://en.wikipedia.org/wiki/R%C3%A9nyi_entropy#R%C3%A9nyi_divergence) of order alpha between the distribution of the mechanism’s outputs on any two datasets that differ in a single element. An (alpha, epsilon)-RDP statement is a relaxation of epsilon-DP but retains many of its important properties that make RDP particularly well-suited for privacy analysis of DP-SGD. The `alphas` parameter instructs the privacy engine what RDP orders to use for tracking privacy expenditure.
 
-When the privacy engine needs to bound the privacy loss of a training run using (epsilon, delta)-DP for a given delta, it searches for the optimal order from among `alphas`. There’s very little additional cost in expanding the list of orders. We suggest using a list `[1 + x / 10.0 for x in range(1, 100)] + list(range(12, 64))`.
+When the privacy engine needs to bound the privacy loss of a training run using (epsilon, delta)-DP for a given delta, it searches for the optimal order from among `alphas`. There’s very little additional cost in expanding the list of orders. We suggest using a list `[1 + x / 10.0 for x in range(1, 100)] + list(range(12, 64))`. You can pass your own alphas by passing `alphas=custom_alphas` when calling `privacy_engine.make_private_with_epsilon`.
 
-A call to `privacy_engine.get_privacy_spent(delta)` returns a pair: an epsilon such that the training run satisfies (epsilon, delta)-DP and an optimal order alpha. An easy diagnostic to determine whether the list of `alphas` ought to be expanded is whether the returned value alpha is one of the two boundary values of `alphas`.
+A call to `privacy_engine.accountant.get_privacy_spent(delta=delta)` returns a pair: an epsilon such that the training run satisfies (epsilon, delta)-DP and an optimal order alpha. An easy diagnostic to determine whether the list of `alphas` ought to be expanded is whether the returned value alpha is one of the two boundary values of `alphas`.
 
 <!-- ## How do I run Opacus in Colab?
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -88,7 +88,7 @@ This statement extends to all downstream uses of this model: its inferences, fin
 
 From the expression above it is obvious that epsilon and delta play different roles: epsilon controls the multiplicative increase in the baseline probability while delta lifts all probabilities by the same amount. For instance, if your baseline scenario (the model trained on *D*′, without your data) assigns 0 probability to some event, the bound on observing this event on *D* (that includes your data) is delta. Because of that, we’d like to target epsilon to be a small constant and select delta to be tiny. A rule of thumb is to set delta to be less than the inverse of the size of the training dataset.
 
-Epsilon and delta are computed *ex post*, following an optimizer run. In fact, for each delta there’s some epsilon, depending on that delta, such that the run satisfies (epsilon, delta)-DP. The call `privacy_engine.accountant.get_privacy_spent(delta=delta)` outputs that epsilon in its first return value.
+Epsilon and delta are computed *ex post*, following an optimizer run. In fact, for each delta there’s some epsilon, depending on that delta, such that the run satisfies (epsilon, delta)-DP. The call `privacy_engine.get_epsilon(delta=delta)` outputs that epsilon in its first return value.
 
 Importantly, (epsilon, delta)-DP is a *conservative upper bound* on the actual privacy loss. There’s [growing](https://arxiv.org/abs/2006.07709) [evidence](https://arxiv.org/pdf/2006.11601.pdf) that the observable privacy loss of the DP-SGD algorithm can be significantly smaller.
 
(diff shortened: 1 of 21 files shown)


opacus/grad_sample/grad_sample_module.py
```diff
@@ -425,6 +425,18 @@ def rearrange_grad_samples(
 
         return activations, backprops
 
+    def state_dict(self, *args, **kwargs):
```
ffuuugor, line 428: Let's reopen this conversation: do we want to reroute *all* unknown (i.e. not explicitly declared in GradSampleModule) attributes to the underlying `_module`?
Right now we only do this for named parameters (see `__getattr__`) .

The reason I'm raising this here is because with new `state_dict` and `load_state_dict` we create a discrepancy between the output of `state_dict` and the output of `named_parameters()` method - which can produce unexpected results.
I think at least we should also override all `named_*` method here too, and at most - think about being more radical and just updating `__getattr__`.

Full context: I was planning to override `named_*` method for other reasons, just didn't get to publishing a PR yet

opacus/privacy_engine.py
```diff
@@ -493,3 +494,50 @@ def get_epsilon(self, delta):
             Privacy budget (epsilon) expended so far.
         """
         return self.accountant.get_epsilon(delta)
+
+    def save_checkpoint(
+        self,
+        *,
+        path: Union[str, os.PathLike, BinaryIO, IO[bytes]],
+        module: GradSampleModule,
+        optimizer: DPOptimizer,
```
ffuuugor, line 503: Should we maybe make optimizer optional here? 

1) You actually only need to save/load optimizer state if you're using parameter scheduling. Otherwise it should be enough to save model and the accountant.

2) Talking about schedulers, what's the recommended way to save/load if you're using noise scheduler? Do we want people to save it by themselves or through privacy engine too?

opacus/tests/accountants_test.py
```diff
@@ -119,3 +119,59 @@ def test_get_noise_multiplier_gdp(self):
...
+        # check loading logic
+        self.assertNotEqual(new_rdp_accountant.state_dict(), accountant.state_dict())
+        new_rdp_accountant.load_state_dict(accountant.state_dict())
+        self.assertEqual(new_rdp_accountant.state_dict(), accountant.state_dict())
+
+        # ensure correct output after completion
+        for _ in range(steps - 1000, steps):
+            new_rdp_accountant.step(
+                noise_multiplier=noise_multiplier, sample_rate=sample_rate
+            )
```
ffuuugor, line 174: That's a great idea for a test!

ffuuugor (commented):
Not surprising at all, but amazing stuff! Thanks for the clean implementation and test coverage!

Slight ask: can we also have a new test in privacy_engine_test for the combined method?

karthikprasad:
Thanks Igor!
> can we also have a new test in privacy_engine_test for the combined method?

Yes, that is still pending, and it the only reason this PR is still in draft mode for this long. I'll add it soon and open it up for review :)

opacus/grad_sample/grad_sample_module.py
karthikprasad: >  do we want to reroute all unknown (i.e. not explicitly declared in GradSampleModule) attributes to the underlying _module?

Good point! Perhaps not! Also, I realise now that we should also store GS specific attributes such as `loss_reduction` and `batch_first` in the `GS.state_dict()`

>  discrepancy between the output of state_dict and the output of named_parameters()

The only difference right now is the extra `_module` prefix, correct?

ffuuugor: > Good point! Perhaps not!

Or perhaps yes...

> I realise now that we should also store GS specific attributes such as loss_reduction and batch_first in the GS.state_dict()

Should we though? Are those fundamentally part of the model state or model architecture?

> The only difference right now is the extra _module prefix, correct?

Affirmative

ffuuugor:
> Yes, that is still pending, and it the only reason this PR is still in draft mode for this long. I'll add it soon and open it up for review :)

Whoops, sorry I jumped the gun - the PR looked so nice and polished, calling to get reviewed

opacus/privacy_engine.py
karthikprasad: Re 1: Point well taken.
Re 2: Scheduler already has a `state_dict()` method. I think it definitely makes sense to save that from Privacy Engine too.

opacus/grad_sample/grad_sample_module.py
karthikprasad: > Should we though? Are those fundamentally part of the model state or model architecture?

They are not, but are essential to instantiate a GS module. I don't see a downside if we do. Do you feel otherwise?

docs/faq.md
```diff
@@ -88,7 +88,7 @@ This statement extends to all downstream uses of this model: its inferences, fin
...
 
-Epsilon and delta are computed *ex post*, following an optimizer run. In fact, for each delta there’s some epsilon, depending on that delta, such that the run satisfies (epsilon, delta)-DP. The call `privacy_engine.get_privacy_spent(delta)` outputs that epsilon in its first return value.
+Epsilon and delta are computed *ex post*, following an optimizer run. In fact, for each delta there’s some epsilon, depending on that delta, such that the run satisfies (epsilon, delta)-DP. The call `privacy_engine.accountant.get_privacy_spent(delta=delta)` outputs that epsilon in its first return value.
```
ffuuugor, line 91: nit: recommended way of accessing epsilon now is `privacy_engine.get_epsilon(delta=delta)`

opacus/grad_sample/grad_sample_module.py
```diff
@@ -425,6 +425,35 @@ def rearrange_grad_samples(
...
         return activations, backprops
 
+    def state_dict(self, *args, **kwargs) -> Dict:
+        """
+        Return the state dict of the wrapped module
+        """
+        ret_state_dict = {
+            f"_module.{key}": value
+            for key, value in self._module.state_dict(*args, **kwargs).items()
+        }
```
ffuuugor, line 435: Are you sure about this?
In my experience "_module" prefix is always an annoyance and never helps. When I was mentioning the discrepancy in my previous comments, I meant that we should approach it from a different end and remove the prefix everywhere. WDYT?

In particular, getting rid of the prefix makes it much easier to load weights from publicly pre-trained models (currently you have to do it strictly before you wrap a model with GSM)

opacus/grad_sample/grad_sample_module.py
```diff
@@ -425,6 +425,35 @@ def rearrange_grad_samples(
...
+        ret_state_dict["batch_first"] = self.batch_first
+        ret_state_dict["loss_reduction"] = self.loss_reduction
```
ffuuugor, line 437: I would vote against having this in the `state_dict`, especially `loss_reduction`.
They are fundamentally part of an architecture, not a state of the model. Additionally, one might want to use old weights stored in a state_dict to train with a different loss, thus changing loss reduction - and I don't think there's anything wrong with that.

`batch_first` if trickier, and I don't have a good use case, but there's always one-to-one mapping between model architecture and batch_first - and we don't store arch in the state_dict

opacus/privacy_engine.py
```diff
@@ -493,3 +495,59 @@ def get_epsilon(self, delta):
...
+        )
+        dict_to_save["privacy_accountant_state_dict"] = self.accountant.state_dict()
+        if optimizer is not None:
+            dict_to_save["optimizer_state_dict"] = optimizer.state_dict()
+        if noise_scheduler is not None:
+            dict_to_save["noise_scheduler_state_dict"] = noise_scheduler.state_dict()
+
+        torch.save(dict_to_save, path, **(torch_save_kwargs or {}))
+
+    def load_checkpoint(
```
ffuuugor, line 531: This handles default values beautifully, thanks for that!

Question: do we want to provide a warning in any of the following scenarios:
1) optimizer is provided, but checkpoint doesn't have optimzier_state_dict
2) checkpoint have optimizer_state_dict, but optimizer isn't provided

opacus/grad_sample/grad_sample_module.py
karthikprasad: > in my previous comments, I meant that we should approach it from a different end and remove the prefix everywhere. WDYT?
Ah! That works too. I don't have a strong preference so long as it is consistent. I'll change that.

> In my experience "_module" prefix is always an annoyance and never helps.

Out of curiosity, what trouble ddi the prefix cause in the past>

opacus/grad_sample/grad_sample_module.py
karthikprasad: Fair points, both!
The way I thought of this is that these two, though not part of a model's state, are necessary to rebuild the model in the same state as before. If a new opacus user wants to save a checkpoint to be reloaded later, they might not know to save these extra variables. However, this can easily be solved with documentation and/or helpful message.

Alternatively, WDYT about saving this is as part of privacy engine checkpoints?

opacus/privacy_engine.py
karthikprasad: I don't see why not.

opacus/grad_sample/grad_sample_module.py
ffuuugor: Scenario 1: You have a "--disable-dp" flag in your pipeline, and you want to train and compare dp and non-dp model. Imagine also, that your infra is unstable, so you checkpoint the model often and run evaluation in a separate job. You then need to handle two different formats of state_dicts
Scenario 2: In general, when you train DP model and then want to transform it to inference model (non-DP obviously), you'd rather have same state_dicts
Scenario 3: When doing public pretraining, today you have to be careful and load_state_dict before you wrap the model with GradSampleModule. With same state_dicts you don't have to worry about it

opacus/grad_sample/grad_sample_module.py
ffuuugor: >  If a new opacus user wants to save a checkpoint to be reloaded later, they might not know to save these extra variables

Do you consider loading your own checkpoint or loading one from a public hub? 

> Alternatively, WDYT about saving this is as part of privacy engine checkpoints?

I'm not sure I see how it can be loaded in a clear way? Would we override parameters set by GradSampleModule constructor for the module that was passed in?

karthikprasad: > Do you consider loading your own checkpoint or loading one from a public hub?

Both.

> I'm not sure I see how it can be loaded in a clear way? Would we override parameters set by GradSampleModule constructor for the module that was passed in?

I agree; it is not clear and also not neat. I do think there is value in saving metadata such as these in the checkpoint, but that would mean having to decide what to do at load time, which is not clear. 

To address this, in the current implementation, I have added support for passing your own dict to `save_checkpoint`. This dict can have all the metadata the user thinks is useful to save along with the state_dicts. At load time, we simply load the states and return the entire loaded dict (state_dicts and metadata) to the user for them to use it the way they see fit.

ffuuugor: > To address this, in the current implementation, I have added support for passing your own dict to save_checkpoint

Oh, that's neat, I love it

opacus/grad_sample/grad_sample_module.py
ffuuugor: JFYI, to clarify things, `state_dict` and `load_state_dict` are already available in GradSampleModule today, as it's inherited from nn.Module, and it behaves exactly like this, so there's no need to override the method even. 

To be fair, one argument for leaving this alone for now is:
a) it doesn't directly help the task at hand, checkpointing can work with either options
b) it will break backward compatibility for some use cases, and even though I do believe it's a change for good, we might want to approach this more carefully.

Before moving forward, let's do one of the following:
1) Decide to deal with parameter names later and remove `state_dict` and `load_state_dict` overrides (because default inherited behaviour is identical)
2) Decide to rename parameters in state_dict now and don't add "_module." prefix

I'm fine with either, and for me that's the only outstanding item before accepting

ffuuugor (approved):
Thanks for the amazing diff and for addressing the comments!
Tests seem to broken in main too, approving

PR #434, opened 2022-05-26
Test Opacus with the latest PyTorch version

romovpa:
Temporarily removing dependency on `torchcsprng` (build is broken) and the corresponding tests to enable integration testing with the latest PyTorch version.

To roll back this after fixing [csprng](https://github.com/pytorch/csprng/) builds.


ffuuugor (approved):
Nice, thanks!

PR #439, opened 2022-06-15
Throws an error when params in optimizer are not the same as that of module's in `make_private`

deepakagrawal:
Summary: Compare nn.Module.parameters() with list of parameters from all param_groups of optimizer. If they are all not equal then raise error "Module parameters are different than optimizer Parameters"

Differential Revision: D37163873

### DIFF
--- a/opacus/privacy_engine.py
+++ b/opacus/privacy_engine.py
@@ -360,6 +360,21 @@ class PrivacyEngine:
         if noise_generator and self.secure_mode:
             raise ValueError("Passing seed is prohibited in secure mode")
 
+        # compare module parameter with optimizer parameters
+        if not all(
+            torch.eq(i, j).all()
+            for i, j in zip(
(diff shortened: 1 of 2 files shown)


karthikprasad (commented):
Thanks Deepak! Looks good. Could you also add a simple unit test in privacy_engine_test for this check?

karthikprasad:
Thanks for adding the tests @deepakagrawal! Please also fix the lint failure and it is good to go after that :)

deepakagrawal:
@karthikprasad can you please check the errors. I think they are not related to the changes I made. I fixed the lint failure.

karthikprasad (approved):
Looks good. Thank you @deepakagrawal!

deepakagrawal:
@karthikprasad pushed the latest commit to github. I will land the diff as well.

PR #440, opened 2022-06-15
Replace torch einsum with opt_einsum

sayanghosh:
Differential Revision: D37128344

### DIFF
--- a/opacus/grad_sample/conv.py
+++ b/opacus/grad_sample/conv.py
@@ -19,6 +19,7 @@ import numpy as np
 import torch
 import torch.nn as nn
 from opacus.utils.tensor_utils import unfold2d, unfold3d
+from opt_einsum import contract
 
 from .utils import register_grad_sampler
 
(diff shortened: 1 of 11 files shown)


sayanghosh:
Optimized einsums obtain significant trends towards an effective reduction in runtime. The most impact is from convolution and linear layers uniformly for all layer sizes. This is under 10 iterations of benchmarking. Runtime_x is from Torch einsums and runtime_y is from opt einsums (in table below).

image
image
image
image

opacus/grad_sample/conv.py
```diff
@@ -19,6 +19,7 @@
 import torch
 import torch.nn as nn
 from opacus.utils.tensor_utils import unfold2d, unfold3d
+from opt_einsum import contract
```
karthikprasad, line 22: `contract` not used in this file?

sayanghosh:
Attached a notebook of the benchmarking and runtime comparison analysis here : https://colab.research.google.com/drive/1WAjSYzyKg7UisShNNwD4FMv6tIMGCzuL#scrollTo=OY0TWxFuXRwt

requirements.txt
```diff
@@ -1,3 +1,4 @@
 numpy>=1.15
 torch>=1.8
 scipy>=1.2
+opt-einsum==3.3.0
```
karthikprasad, line 4:
```suggestion
opt-einsum>=3.3.0
```

karthikprasad (approved):
Thanks. Looks good to me. Please fix the lint error before merging.

PR #441, opened 2022-06-17
Experimental expanded weights

ffuuugor:


opacus/grad_sample/gsm_exp_weights.py
```diff
@@ -0,0 +1,35 @@
...
+        try:
+            from torch.nn.utils._per_sample_grad import call_for_per_sample_grads
+            self.call_for_per_sample_grads = call_for_per_sample_grads
+        except ImportError:
+            raise ImportError(
+                f"Requested grad_sample_mode=ew, "
+                f"but found PyTorch version={torch.__version__}. "
+                f"ExpandedWeights available for torch>=1.12. "
+                f"Please install recent PyTorch or use grad_sample_mode=hooks"
+            )
```
ffuuugor, line 26: NB: this will allow importing the module, but fail on initialization for older PyTorch environments

opacus/optimizers/optimizer.py
```diff
@@ -285,10 +245,56 @@ def __init__(
...
+            ret = p.grad_sample
+        elif isinstance(p.grad_sample, list):
+            ret = torch.cat(p.grad_sample, dim=0)
+        else:
+            raise ValueError(f"Unexpected grad_sample type: {type(p.grad_sample)}")
+
+        if self.ew_compatibility_mode and self.loss_reduction == "mean":
+            return ret * len(ret)
+        else:
+            return ret
```
ffuuugor, line 296: This is required because ExpandedWeights always assume loss_reduction="sum"

opacus/optimizers/optimizer.py
```diff
@@ -393,7 +399,7 @@ def clip_and_accumulate(self):
         """
 
         per_param_norms = [
-            g.view(len(g), -1).norm(2, dim=-1) for g in self.grad_samples
+            g.reshape(len(g), -1).norm(2, dim=-1) for g in self.grad_samples
```
ffuuugor, line 402: This has to do with some tensor memory allocation details specific to EW. The way I understand it, `view` can only work with contiguous inputs, which is not the case for EW grads

opacus/tests/privacy_engine_test.py
```diff
@@ -839,22 +871,27 @@ def __init__(self, tie=True):
...
             torch.randint(low=0, high=100, size=(self.DATA_SIZE,)),
         )
-        return DataLoader(ds, batch_size=self.BATCH_SIZE, drop_last=True)
+        return DataLoader(ds, batch_size=self.BATCH_SIZE, drop_last=False)
 
     def _init_model(
         self, private=False, state_dict=None, model=None, **privacy_engine_kwargs
     ):
-        return SampleTiedWeights(tie=False)
+        return SampleTiedWeights(tie=True)
```
ffuuugor, line 897: Fun fact: this was broken for a while

opacus/tests/privacy_engine_test.py
```diff
@@ -686,7 +707,7 @@ def __init__(self):
         self.conv2 = nn.Conv1d(16, 32, 3, 1)
         self.lnorm1 = nn.LayerNorm((32, 23))
         self.conv3 = nn.Conv1d(32, 32, 3, 1, bias=False)
-        self.instnorm1 = nn.InstanceNorm1d(32, affine=True)
+        # self.instnorm1 = nn.InstanceNorm1d(32, affine=True)
```
ffuuugor, line 710: temporary fix, see https://github.com/pytorch/pytorch/pull/79800

opacus/tests/privacy_engine_test.py
```diff
@@ -196,117 +210,89 @@ def _compare_to_vanilla(self, do_noise, do_clip, expected_match, use_closure=Fal
...
-                        expected_match=not (do_noise or do_clip),
-                    )
+        self._compare_to_vanilla(
+            do_noise=do_noise,
+            do_clip=do_clip,
+            expected_match=not (do_noise or do_clip),
+            use_closure=use_closure,
+            grad_sample_mode=self.GRAD_SAMPLE_MODE,
+            max_steps=max_steps,
+        )
```
ffuuugor, line 272: Fun fact: this was broken for a while too (we're supposed to call `self._compare_to_vanilla_accumulated`, which is now refactored into `self._compare_to_vanilla` with extra `max_steps` parameter)

opacus/grad_sample/README.md
```diff
@@ -1,16 +1,56 @@
...
+| Required PyTorch version | 1.8+ | 1.12+ |
+| Development status | Underlying mechanism deprecated | Beta |
+| Performance | - | ✅ Likely up to 2.5x faster |
+| torchscript models | Not supported | ✅ Supported |
+| Client-provided grad sampler | ✅ Supported | Not supported |
+| `batch_first=False` | ✅ Supported | Not supported |
+| Most popular nn.* layers | ✅ Supported | ✅ Supported |
+| Recurrent networks | ✅ Supported | Not supported |
+| nn.LayerNorm | ✅ Supported | Not supported |
+| nn.Conv3d | ✅ Supported | Not supported |
```
alexandresablayrolles, line 56: Are LayerNorm and nn.Conv3d *not* supported? From the latest pytorch nightly it seems that they are:
```python
from torch.nn.utils._expanded_weights import ExpandedWeight
print(ExpandedWeight.handled_functions)
# {<built-in method conv3d of type object at 0x7fe1208c50c0>: <class 'torch.nn.utils._expanded_weights.conv_expanded_weights.ConvPerSampleGrad'>,  <function layer_norm at 0x7fe0cb23f1f0>: <class 'torch.nn.utils._expanded_weights.layer_norm_expanded_weights.LayerNormPerSampleGrad'>, ...}
```

opacus/grad_sample/README.md
```diff
@@ -1,16 +1,56 @@
...
+| xxx | Hooks | Expanded Weights |
+|:-----:|:-------:|:------------------:|
+| Required PyTorch version | 1.8+ | 1.12+ |
+| Development status | Underlying mechanism deprecated | Beta |
+| Performance | - | ✅ Likely up to 2.5x faster |
+| torchscript models | Not supported | ✅ Supported |
+| Client-provided grad sampler | ✅ Supported | Not supported |
+| `batch_first=False` | ✅ Supported | Not supported |
+| Most popular nn.* layers | ✅ Supported | ✅ Supported |
+| Recurrent networks | ✅ Supported | Not supported |
```
alexandresablayrolles, line 54: Out of curiosity, what is the problem with using our "DPRNN" with ExpandedWeights? I'm guessing that they forbid accumulation of grad_sample?

opacus/optimizers/optimizer.py
alexandresablayrolles: I think we can skip the reshape and directly compute the norm as:
```python
g.norm(2, dim=tuple(range(1, g.ndim)))
```

opacus/optimizers/perlayeroptimizer.py
```diff
@@ -51,13 +52,16 @@ def __init__(
...
             secure_mode=secure_mode,
+            ew_compatibility_mode=ew_compatibility_mode,
         )
 
     def clip_and_accumulate(self):
         for (p, max_grad_norm) in zip(self.params, self.max_grad_norms):
             _check_processed_flag(p.grad_sample)
 
-            per_sample_norms = p.grad_sample.view(len(p.grad_sample), -1).norm(2, dim=1)
+            per_sample_norms = p.grad_sample.reshape(len(p.grad_sample), -1).norm(
```
alexandresablayrolles, line 62: Same here: no need for reshape before computing norm.

alexandresablayrolles (approved):
Impressive diff! I like where this is going. Couple of nits here and there but overall LGTM.

opacus/grad_sample/README.md
ffuuugor: See [colab](https://colab.research.google.com/drive/1ihar3_rc_P5otxRL6xP929BxjuiZRi7A#scrollTo=iwOrjL_kfFwG) for details. They have corresponding grad sample methods, but they don't seem to work properly and don't pass our tests.

1) LayerNorm - I'll switch the label to "Unstable" rather than "not supported". It throws and exception if it's the only layer in the model (quite unrealistic) and otherwise seems to work fine.
2) Conv3d - Throws shape error

opacus/grad_sample/README.md
ffuuugor: I'll give you a quote from @samdow which I do not claim to fully understand:

>Backstory here: the reason RNNs are difficult for a lot of people is because the kernel will also return a buffer that's used by cuda. This means we can't write a custom per sample gradients formula because we don't have any information as to what is in the buffer and what the kernels are using it for

> So this makes it weird for a lot of systems to trace because they're returning this buffer object that should never really be user facing. What we hope to do is add a flag to the RNN module that allows a user to choose to basically decompose the RNN into a linear layer (like what Opacus has). This will make it easier to have interpretable traces and let researchers experiment on something that they understand (instead of something that returns a buffer)

opacus/optimizers/perlayeroptimizer.py
ffuuugor: Indeed, will update, thanks!

opacus/grad_sample/README.md
alexandresablayrolles: Well, so it doesn't really say what would happen with a decomposed RNN like Opacus has. Let's try!

ffuuugor: huh, good point

opacus/grad_sample/grad_sample_module.py
```diff
@@ -73,7 +72,7 @@ def promote_current_grad_sample(p: nn.Parameter) -> None:
         del p._current_grad_sample
 
 
-class GradSampleModule(nn.Module):
+class GradSampleModule(AbstractGradSampleModule):
     r"""
     Extends nn.Module so that its parameter tensors have an extra field called .grad_sample.
```
karthikprasad, line 77: nit: rectify documentation

opacus/privacy_engine.py
```diff
@@ -203,28 +212,33 @@ def _prepare_model(
         *,
         batch_first: bool = True,
         loss_reduction: str = "mean",
-    ) -> GradSampleModule:
+        grad_sample_mode: str = "hooks",
```
karthikprasad, line 215: How do you feel about making subclasses of PrivacyEngine and attaching it to one kind of GSM? Say `PrivacyEngine` and `PrivacyEngineWithExpandedWeights`?

Pros: (1) avoids pesky `if-else` (2) makes it easier and clearer to the end user and developer to follow since one class only has to deal with one implementation (3) extensible: we can have `PrivacyEngineWithFuncTorch` in the future if needed without modifying the existing code.

Cons: (1) More user facing classes (2) unnecessary if we want to eventually deprecate one of the GSM approaches. (3) <I'm sure I'm missing some, but can't think of any; please comment with more if you see it>

opacus/grad_sample/README.md
ffuuugor: So yeah, our custom DPRNN doesn't work too. If fails in a funny way - EW just doesn't compute `grad_sample` without throwing any exceptions.
This is clearly not a desired behaviour and I've included that into a [collab](https://colab.research.google.com/drive/1ihar3_rc_P5otxRL6xP929BxjuiZRi7A#scrollTo=RdiMJNRoWKsm) documenting required improvements

opacus/privacy_engine.py
ffuuugor: To be honest, I'm not a big fan.

At this point the only customisation required to switch between EW and hooks is the class name, and I think PrivacyEngine already acts like sort of factory, mapping parameters (e.g. `distributed` or `clipping`) to the most appropriate implementation.

>  avoids pesky if-else

True, but I don't see it as a big nuisance

> makes it easier and clearer to the end user and developer to follow since one class only has to deal with one implementation

I would argue, that controlling behaviour with a string-argument is an accepted practice, and for most users `grad_sample_mode` is an insignificant implementation detail. And if you face an issue with the default implementation, would it be easier for you to pass an extra argument or switch to a different class?

> extensible: we can have PrivacyEngineWithFuncTorch in the future if needed without modifying the existing code.

I don't think we have good enough picture on how true-functorch opacus would look like. I agree it's a possible future, but I'm struggling to take this into account due to amount of ambiguity

 > More user facing classes

++

> unnecessary if we want to eventually deprecate one of the GSM approaches.

It's a good question whether we will. Ideally, we'd want to, but it does feel like we'd have to live with both for a while

>  <I'm sure I'm missing some, but can't think of any; please comment with more if you see it>

For me the strongest point is that with `grad_sample_mode` as an argument we have better control over default behaviour. For example at some point we'll be able to switch ExpandedWeights by default when they reach certain point of maturity

karthikprasad: I'm not entirely convinced, but I think you have said enough to make me not block you :)

My annoyance is largely due to this:
> I would argue, that controlling behaviour with a string-argument is an accepted practice, and for most users grad_sample_mode is an insignificant implementation detail. And if you face an issue with the default implementation, would it be easier for you to pass an extra argument or switch to a different class?

1. I don't think this is an insignificant implementation detail. If the difference were just one being faster than the other, I would have agreed, but here, the two implementations differ in what they support too. This, to me, is a huge change.
2.  I typically think of flags as changing a behaviour of an execution flow as it is flowing, not changing the whole component. Eg, accountants in opacus right now are different classes, and rightly so. It could have easily been an `if-else` gated logic enabled by a `string` flag, but that wouldn't have been clean.


However, as I said above, my annoyance is not terrible enough to block the PR.

ffuuugor: I would gladly continue the discussion and not rush the PR until we find a common ground.

> Eg, accountants in opacus right now are different classes, and rightly so.

Buuut, accountants are controlled with a string argument (see `PrivacyEngine` constructor). They are implemented as different classes, but exposed to the client through factory-like behaviour in `PrivacyEngine`

I agree that `GradSampleModule` implementation is important and does potentially change the behaviour. Perhaps my argument on it being merely an implementation detail doesn't hold water and I withdraw it.

However, this doesn't change my position on subclassing PrivacyEngine. If we go back to 1.0 designs, one of the pain points for us was the combinatorial explosion from having multiple optimizers and accountants. If we were to use `PrivacyEngine` subclasses, we'd need to create at least 16 of them: 2 accountants, 2 clipping strategies, disrtibuted/non-distributed, poisson/non-poisson (and now different grad samplers).

We've played with it in PR #254, but it ended up being too cumbersome: navigating the code became too hard.

And I don't think it make sense to have new `PrivacyEngine` subclasses for different GradSampleModules, but not optimizers or data loaders

opacus/grad_sample/README.md
samdow: I can shine some light on this! So the more generic mechanism that EW is using is called reparametrization and it assumes that the names of the parameters match what PyTorch would create. At a high level, because the DPRNN layers are renaming their parameters, it terminates the reparameterization algorithm early so we skip turning the original weights into expanded weights (meaning that they don't get the grad_sample attribute, yikes!). This is probably going to be difficult to error on because it's hard to detect when a user is changing the names of parameters.

The reason that changing the RNN behavior with a flag should work is because it will still be the RNN layer, we'll just change what's happening in the forward pass if the flag is set. I assumed that the renaming trick was so that exporting models worked correctly so that we couldn't this behavior in Opacus, but without the renaming it should work with EW (yet not with exporting 😞 )

opacus/privacy_engine.py
karthikprasad: > Buuut, accountants are controlled with a string argument (see PrivacyEngine constructor). They are implemented as different classes, but exposed to the client through factory-like behaviour in PrivacyEngine

Touché. A small nit in that case, accept the gsm_mode as an arg to `PrivacyEngine` rather than `make_private`. To follow the same pattern?

karthikprasad (approved):

PR #445, opened 2022-07-04
[RFC] Functorch to automatically compute per-sample gradients

alexandresablayrolles:
**TL;DR** We can use Functorch to compute per sample gradients automatically.

This PR proposes to use Functorch to compute per sample gradients. The "meat" is in grad_sample/functorch.py. It basically does a mini-"per sample" forward/backward on the layer at hand, with the tricks that outputs.backward(backprops) is equivalent to (outputs * backprops).sum().backward().

## Open questions

1. How do we validate the modules? It used to be clear that any module which grad_sampler was not implemented was not OK but now it's different. Should we move from a model where everything not allowed was forbidden, to a model where everything not forbidden is allowed.
2. Do we want to provide clean support for activation/de-activation of GRAD_SAMPLERS or is it ok to just del GRAD_SAMPLERS[module]?

### DIFF
--- a/opacus/grad_sample/functorch.py
+++ b/opacus/grad_sample/functorch.py
@@ -1,0 +1,60 @@
+from functorch import grad, make_functional, vmap
+from opacus.layers.dp_rnn import RNNLinear
+
+
+def prepare_layer(layer, batch_first=True):
+    """
+    Prepare a layer to compute grad samples using functorch.
+    The grad samples are computed by redoing the forward and
+    backward passes on the functional version of the module.
+
+    Args:
+        layer: the layer to prepare
+        batch_first: whether the input is batch_first or not
+    """
+    if len(list(layer.buffers())) > 0:
+        raise NotImplementedError(
+            "This layer has buffers and is not supported by Opacus"
+        )
+    flayer, _ = make_functional(layer)
+
+    def compute_loss_stateless_model(params, activations, backprops):
+        if batch_first or type(layer) is RNNLinear:
+            batched_activations = activations.unsqueeze(0)
+            batched_backprops = backprops.unsqueeze(0)
+        else:
+            # If batch_first is False, the batch dimension is the second dimension
+            batched_activations = activations.unsqueeze(1)
+            batched_backprops = backprops.unsqueeze(1)
+
+        output = flayer(params, batched_activations)
+        loss = (output * batched_backprops).sum()
+
+        return loss
+
+    ft_compute_grad = grad(compute_loss_stateless_model)
+    # Note that the vmap is done on the first dimension, regardless of batch_first
+    # This is because the activations and backprops given by the GradSampleModule
+    # are always batch_first=True
+    layer.ft_compute_sample_grad = vmap(ft_compute_grad, in_dims=(None, 0, 0))
+
+
+def ft_compute_per_sample_gradient(layer, activations, backprops):
+    """
+    Compute the per-sample gradient of the layer.
+    Args:
+        layer: the layer on which to compute the gradient
+        activations: the input to the layer
+        backprops: the  gradient of the loss w.r.t. outputs of the layer
+    """
+    parameters = list(layer.parameters())
+    if not hasattr(layer, "ft_compute_sample_grad"):
+        prepare_layer(layer)
+
+    per_sample_grads = layer.ft_compute_sample_grad(parameters, activations, backprops)
+
+    ret = {}
+    for i_p, p in enumerate(parameters):
+        ret[p] = per_sample_grads[i_p]
+
+    return ret
--- a/opacus/grad_sample/functorch.py
+++ b/opacus/grad_sample/functorch.py
@@ -1,4 +1,3 @@
-from functorch import grad, make_functional, vmap
 from opacus.layers.dp_rnn import RNNLinear
 
 
@@ -12,6 +11,8 @@ def prepare_layer(layer, batch_first=True):
         layer: the layer to prepare
         batch_first: whether the input is batch_first or not
     """
+    from functorch import grad, make_functional, vmap
+
     if len(list(layer.buffers())) > 0:
         raise NotImplementedError(
             "This layer has buffers and is not supported by Opacus"
--- a/opacus/grad_sample/grad_sample_module.py
+++ b/opacus/grad_sample/grad_sample_module.py
@@ -16,18 +16,21 @@
 from __future__ import annotations
 
 import logging
+import warnings
 from functools import partial
 from typing import List, Tuple
 
 import torch
 import torch.nn as nn
 from opacus.grad_sample.gsm_base import AbstractGradSampleModule
-from opacus.layers.dp_rnn import DPRNNBase, DPRNNCellBase, RNNLinear
+from opacus.layers.dp_rnn import DPGRU, DPLSTM, DPRNN, RNNLinear
 from opacus.utils.module_utils import (
     requires_grad,
     trainable_modules,
     trainable_parameters,
 )
+
+from .functorch import ft_compute_per_sample_gradient, prepare_layer
 
 
 logger = logging.getLogger(__name__)
@@ -89,6 +92,7 @@ class GradSampleModule(AbstractGradSampleModule):
         batch_first=True,
         loss_reduction="mean",
         strict: bool = True,
+        force_functorch=False,
     ):
         """
 
@@ -108,6 +112,9 @@ class GradSampleModule(AbstractGradSampleModule):
                 possible and set to None otherwise. This is not recommended, because
                 some unsupported modules (e.g. BatchNorm) affect other parameters and
                 invalidate the concept of per sample gradients for the entire model.
+            force_functorch: If set to ``True``, will use functorch to compute
+                all per sample gradients. Otherwise, functorch will be used only
+                for layers without registered grad sampler methods.
 
         Raises:
             NotImplementedError
@@ -128,13 +135,24 @@ class GradSampleModule(AbstractGradSampleModule):
             )
 
         self.hooks_enabled = False
-        self.add_hooks(loss_reduction=loss_reduction, batch_first=batch_first)
+        self.batch_first = batch_first
+        self.loss_reduction = loss_reduction
+        self.force_functorch = force_functorch
+        self.add_hooks(
+            loss_reduction=loss_reduction,
+            batch_first=batch_first,
+            force_functorch=force_functorch,
+        )
 
     def forward(self, *args, **kwargs):
         return self._module(*args, **kwargs)
 
     def add_hooks(
-        self, *, loss_reduction: str = "mean", batch_first: bool = True
+        self,
+        *,
+        loss_reduction: str = "mean",
+        batch_first: bool = True,
+        force_functorch: bool = False,
     ) -> None:
         """
         Adds hooks to model to save activations and backprop values.
@@ -151,6 +169,8 @@ class GradSampleModule(AbstractGradSampleModule):
                 ``[K, batch_size, ...]``
             loss_reduction: Indicates if the loss reduction (for aggregating the gradients)
                 is a sum or a mean operation. Can take values "sum" or "mean"
+            force_functorch: If set to ``True``, will use functorch to compute all per sample gradients.
+                Otherwise, functorch will be used only for layers without registered grad sampler methods.
         """
         if hasattr(self._module, "autograd_grad_sample_hooks"):
             raise ValueError("Trying to add hooks twice to the same model")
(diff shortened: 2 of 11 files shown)


opacus/tests/grad_samples/conv1d_test.py
```diff
@@ -48,7 +49,8 @@ def test_conv1d(
         dilation: int,
         groups: int,
     ):
-
+        if nn.Conv1d in GradSampleModule.GRAD_SAMPLERS:
```
ffuuugor, line 52: huh?

opacus/grad_sample/functorch.py
```diff
@@ -0,0 +1,28 @@
...
+
+        return loss
+
+    ft_compute_grad = grad(compute_loss_stateless_model)
+
+    layer.ft_compute_sample_grad = vmap(ft_compute_grad, in_dims=(None, 0, 0))
+    layer.func_params = params
+
+
+def compute_per_sample_gradient(layer, activations, backprops):
```
ffuuugor, line 21: nit: I'd call it smth like `ft_compute_per_sample_gradient` or otherwise that would indicate that it's functorch-based functionality.

And probably explicitly check that `prepare_layer` has been called on the input `layer`

opacus/grad_sample/grad_sample_module.py
```diff
@@ -219,20 +220,22 @@ def add_hooks(
             self.autograd_grad_sample_hooks = self._module.autograd_grad_sample_hooks
 
         for _module_name, module in trainable_modules(self._module):
-            if type(module) in self.GRAD_SAMPLERS:
-                self.autograd_grad_sample_hooks.append(
-                    module.register_forward_hook(self.capture_activations_hook)
-                )
+            if not type(module) in self.GRAD_SAMPLERS: 
+                prepare_layer(module)
```
ffuuugor, line 224: question: can we call `prepare_layer` from within `compute_per_sample_gradient`? To avoid extra preparation step

ffuuugor: Oh, and we probably want to add extra bits to '_close()' method to remove extra attributes we add in `prepare_layer`

ffuuugor (commented):
That's amazing! a) the fact that it works and b) the fact that it looks so neat and elegant is just great.

One question I have about making this a part of classic `GradSampleModule` - are there any edge cases where functorch could do things wrong? Like RNNs, tied weights, GANs, etc (that's just a list of things I remember causing issues in the past, I don't necessarily see a reason they shouldn't work with functorch)

> How do we validate the modules?

I think what you did there is good as a first step. We don't have to solve all of our problems at once, and the problem of custom layers doing strange transforms existed before. 

I think in the future we can provide a utility function that mimics our grad_sample tests to give clients an opportunity to check their models, but for now I think we shouldn't try to kill two birds with one stone.

> Do we want to provide clean support for activation/de-activation of GRAD_SAMPLERS or is it ok to just del GRAD_SAMPLERS[module]?

What could be the reason for a client to disable custom grad_sampler?

opacus/grad_sample/grad_sample_module.py
alexandresablayrolles: > question: can we call `prepare_layer` from within `compute_per_sample_gradient`? To avoid extra preparation step

Sounds a bit wasteful no? Especially since GSM already does a preparation step to add hooks.

ffuuugor (commented):
Looks great!
Changes on `GradSampleModule` level LGTM, now let's make them available in `PrivacyEngine` API. I think it's best to reuse 'grad_sample_mode' parameter for that, adding the 3rd possible value there.

Regarding the tests,
1) For the "module not supported" tests in `tests/grad_sample_module_test.py` I would just switch False to True - we want to make sure GSM doesn't raise a validation exception for the unknown modules now
2) I think it's a good idea to update `tests/gradient_accumulation_test.py` as well. I'm not sure what behaviour functorch provides, but we want to know it and document it.
3) `tests/module_validator_test.py` should stay as it is, we still don't want BatchNorm
4) It'd be good to add another use case to `tests/privacy_engine_test.py` covering custom modules and non-supported core modules (maybe something like `nn.Transformer*`)

Again, amazing diff, really a step forward for opacus

opacus/grad_sample/grad_sample_module.py
```diff
@@ -124,7 +129,12 @@ def __init__(
         self.hooks_enabled = False
         self.batch_first = batch_first
         self.loss_reduction = loss_reduction
-        self.add_hooks(loss_reduction=loss_reduction, batch_first=batch_first)
+        self.add_hooks(
+            loss_reduction=loss_reduction,
+            batch_first=batch_first,
+            force_functorch=force_functorch,
+        )
+        self.force_functorch = force_functorch
```
karthikprasad, line 137: teeny tiny nit: move this line below L131

opacus/grad_sample/grad_sample_module.py
```diff
@@ -257,6 +275,11 @@ def remove_hooks(self) -> None:
             delattr(self, "autograd_grad_sample_hooks")
             delattr(self._module, "autograd_grad_sample_hooks")
 
+        # Remove functorch hooks
+        for _module_name, module in trainable_modules(self._module):
+            if not type(module) in self.GRAD_SAMPLERS:
+                delattr(module, "ft_compute_sample_grad")
```
karthikprasad, line 281: Shouldn't this be done for all layers when force_functorch is True?

opacus/grad_sample/functorch.py
```diff
@@ -0,0 +1,38 @@
+from functorch import make_functional, vmap, grad
+from opacus.layers.dp_rnn import RNNLinear
+
+
+def prepare_layer(layer, batch_first=True):
```
karthikprasad, line 5: Please add documentation explaining what this function does.

opacus/grad_sample/grad_sample_module.py
```diff
@@ -444,9 +471,10 @@ def is_supported(cls, module: nn.Module) -> bool:
         Returns:
             ``True`` if grad sampler is found, ``False`` otherwise
         """
-        return type(module) in cls.GRAD_SAMPLERS or isinstance(
-            module, (DPRNNBase, DPRNNCellBase)
-        )
+        return True
+        # return type(module) in cls.GRAD_SAMPLERS or isinstance(
+        #     module, (DPRNNBase, DPRNNCellBase)
+        # )
```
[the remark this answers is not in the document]
karthikprasad: 🦕

opacus/layers/dp_rnn.py
```diff
@@ -630,6 +630,24 @@ def initialize_cells(self):
         self.set_rename_map(rename_map)
         return cells
 
+    def parameters(self, recurse=True):
+        """
+        Redefining parameters() as the RenameParamsMixin confuses it
```
karthikprasad, line 635: sorry, confuses what?

opacus/tests/grad_samples/conv1d_test.py
```diff
@@ -21,6 +21,7 @@
 from hypothesis import given, settings
 
 from .common import GradSampleHooks_test, expander, shrinker
+from opacus.grad_sample import GradSampleModule
```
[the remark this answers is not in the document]
karthikprasad: 🦖

requirements.txt
```diff
@@ -1,3 +1,4 @@
+functorch
```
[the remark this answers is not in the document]
karthikprasad: The main thing that held us back from bumping up the torch dependency was csprng. Sorry, I'm not familiar with the latest take on that, are we okay with opacus sans csprng?

opacus/grad_sample/functorch.py
```diff
@@ -0,0 +1,38 @@
...
+        output = flayer(params, batched_activations)
+        loss = (output * batched_backprops).sum()
+
+        return loss
+
+    ft_compute_grad = grad(compute_loss_stateless_model)
+    layer.ft_compute_sample_grad = vmap(ft_compute_grad, in_dims=(None, 0, 0))
+
+
+def ft_compute_per_sample_gradient(layer, activations, backprops):
```
karthikprasad, line 27: ditto

opacus/grad_sample/functorch.py
```diff
@@ -0,0 +1,38 @@
+from functorch import make_functional, vmap, grad
```
karthikprasad, line 1: Wow! So elegant! <3

opacus/grad_sample/grad_sample_module.py
```diff
@@ -211,6 +225,7 @@ def add_hooks(
                 ``[K, batch_size, ...]``
             loss_reduction: Indicates if the loss reduction (for aggregating the gradients)
                 is a sum or a mean operation. Can take values "sum" or "mean"
+            force_functorch: If set to ``True``, will use functorch to compute per sample gradients.
```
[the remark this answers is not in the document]
karthikprasad: Also worth explaining, either here or on the website/readme, as to why one should choose one over the other.

opacus/grad_sample/grad_sample_module.py
```diff
@@ -352,7 +375,11 @@ def capture_backprops_hook(
             loss_reduction=loss_reduction,
             batch_first=batch_first,
         )
-        grad_sampler_fn = self.GRAD_SAMPLERS[type(module)]
+        if not self.force_functorch and type(module) in self.GRAD_SAMPLERS:
+            grad_sampler_fn = self.GRAD_SAMPLERS[type(module)]
+        else:
+            grad_sampler_fn = ft_compute_per_sample_gradient
+
```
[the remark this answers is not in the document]
alexandresablayrolles: It's all in `rearrange_grad_samples` that puts the batch first if needed and scales backprops appropriately for `loss_reduction`

requirements.txt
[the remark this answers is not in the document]
alexandresablayrolles: > are we okay with opacus sans csprng?

csprng has been an optional dependency since fc23b23ed19a9d0bdc1d1a7a55bcef3a5904b36f (which means you could have functorch+pytorch1.11 without csprng even though it's not ideal). That being said, I think @romovpa fixed the csprng dependency on older version of Pytorch but not 100% sure.

> given that functorch requires torch>=1.11, do we want to update our dependency too?

If we put functorch, wouldn't pip automatically apply this constraint to the torch version?

opacus/layers/dp_rnn.py
alexandresablayrolles: For some reason, RenameParamsMixin gives parameters() for both the DPRNN and the RNNLinear, which are the same. This is a way around it but there might be something cleaner?

opacus/grad_sample/grad_sample_module.py
alexandresablayrolles: Done. I'm thinking, we can also just check dynamically if the attribute is there and remove it?

karthikprasad: Thanks. That works too.

requirements.txt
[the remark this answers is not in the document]
karthikprasad: > csprng has been an optional dependency since fc23b23

That's right! IIUC, csprng is optional dependency, but it can still be used if needed. csprng is only supported until torch 1.8, and the current opacus is compatible with torch 1.8.
However, if we change the torch version from `torch>=1.8` to `torch>=1.11` in L3 in this file, opacus will no longer be useable with torch 1.8 and consequently with csprng.
I agree that @romovpa would have the best context here.

opacus/layers/dp_rnn.py
karthikprasad: I don't know if there is a cleaner way; this implementation is clean enough TBH. My question was mostly around what "it" refers to in the docstring; elaborating it would be useful :)

alexandresablayrolles: I found a way around that (basically not triggering grad samples on DPRNN/DPLSTM as it is done on their LSTMLinear), not needed anymore!

ffuuugor (approved):
Love this!
Approving, as all my comments are small nits.

One more thing we want to have is additional test in "tests/privacy_engine_test.py" to confirm that functorch produces correct results.
It doesn't have to be a part of this PR - can you pls create a new issue so that we follow up on this later?

opacus/grad_sample/grad_sample_module.py
```diff
@@ -374,10 +409,13 @@ def is_supported(cls, module: nn.Module) -> bool:
         Returns:
             ``True`` if grad sampler is found, ``False`` otherwise
         """
-        return type(module) in cls.GRAD_SAMPLERS or isinstance(
-            module, (DPRNNBase, DPRNNCellBase)
+        warnings.warn(
+            "GradSampleModule.is_supported is deprecated, as all layers can now be used with functorch.",
```
karthikprasad, line 413: The biggest impacc of this change! Thank you

requirements.txt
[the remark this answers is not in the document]
karthikprasad: Just reopened this thread. Did we reach a conclusion on which torch version to support?
Given that functorch is listed as a requirement, what would happen with when someone tries to use functorch on torch 1.8.1?

karthikprasad (approved):

PR #451, opened 2022-07-13
Add support for 'same' and 'valid' for padding in conv

ashkan-software:
Differential Revision: D37811390

### DIFF
--- a/opacus/tests/grad_samples/conv2d_test.py
+++ b/opacus/tests/grad_samples/conv2d_test.py
@@ -22,7 +22,7 @@ from hypothesis import given, settings
 from opacus.utils.tensor_utils import unfold2d
 from torch.testing import assert_allclose
 
-from .common import GradSampleHooks_test, expander, shrinker
+from .common import expander, GradSampleHooks_test, shrinker
 
 
 class Conv2d_test(GradSampleHooks_test):
@@ -34,7 +34,7 @@ class Conv2d_test(GradSampleHooks_test):
         out_channels_mapper=st.sampled_from([expander, shrinker]),
         kernel_size=st.integers(2, 3),
         stride=st.integers(1, 2),
-        padding=st.sampled_from([0, 2]),
+        padding=st.sampled_from([0, 2, 'same', 'valid']),
         dilation=st.integers(1, 3),
         groups=st.integers(1, 16),
     )
@@ -52,7 +52,8 @@ class Conv2d_test(GradSampleHooks_test):
         dilation: int,
         groups: int,
     ):
-
+        if (padding == 'same' and stride != 1):
+            return
         out_channels = out_channels_mapper(C)
         if (
             C % groups != 0 or out_channels % groups != 0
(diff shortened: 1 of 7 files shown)


ffuuugor:
So the reason for the failing tests in that `padding="same"` introduces uneven padding, which previously wasn't supported by either PyTorch or opacus.

All numerical values of padding in convolution layers pad equally from both sides, which we assumed to be the case when writing grad sampler for opacus.
With padding="same" it's possible to have uneven padding: if total padding length is odd number, the padding on the right will be larger by one.

I've prototyped the solution in #456 - it's based on your commit, I've fixed 1d and 2d cases. Still pending is:
1) Fix 3d case (similar to 2d)
2) Run benchmarks. I call for F.pad() in all cases, regardless of the input parameters. This is the most straightforward option, but might not be the most efficient. It's possible that doing padding inside conv layer is more efficient - if that's true, we'd want to add an extra if-statement and only do padding manually if required (`F.pad` allows uneven padding, conv layer doesn't)
3) Ideally, we'd like this to be supported by ExpandedWeights too. So we'd need to talk to @samdow about adapting this changes in core pytorch code

ffuuugor:
@ashkan-software do you know why `commit` CircleCI tests weren't triggered?
LGTM overall, I can accept if all the tests and linters pass

ffuuugor:
@ashkan-software For ExpandedWeights tests failures, I don't think they should be blocking for merging this. For now you can disable this particular combination of input arguments (we have `ew_compatible` parameter in `_run` method for this).
And add a `#TODO` to the code so we remember to enable it later when EW starts supporting it.
Additionally, it'd be great if you add this info to `grad_sample/README.md`, where we document discrepancies between hook-based and ew-based GradSampleModules.

@samdow, do we have an issue created for this, so we can link it for future tracking?

ashkan-software:
@ffuuugor thanks a lot for your recommendations. I applied them all and we are good to go!

samdow:
> [@samdow](https://github.com/samdow), do we have an issue created for this, so we can link it for future tracking?

No but I'm planning to add this today thanks to talk with @ashkan-software! If I don't get to it today, I'll file an issue 😄 Thanks for flagging this and doing the heavy lifting of the implementation

opacus/tests/grad_samples/conv2d_test.py
```diff
@@ -69,21 +70,29 @@ def test_conv2d(
             dilation=dilation,
             groups=groups,
         )
-        self.run_test(x, conv, batch_first=True, atol=10e-5, rtol=10e-4)
+        is_ew_compatible = padding != 'same'  # TODO add support for padding = 'same' with EW
```
ffuuugor, line 73: does "valid" padding work fine with EW?

ffuuugor (approved):
Looks great, thanks a lot for the amazing work!

opacus/tests/grad_samples/conv2d_test.py
ashkan-software: Good question. It should since it's basically zero padding. But let me run a test locally and verify this. Good catch!

ashkan-software: well, actually on line 37 of this file we have unit tests for both "valid" and "same". So it should work just fine 😎

PR #453, opened 2022-07-19
new ExpandedWeights API

ffuuugor:
as ExpandedWeights are still in beta phase, breaking changes are expected (e.g. [#80892](https://github.com/pytorch/pytorch/pull/80892)).

See also [test failures](https://app.circleci.com/pipelines/github/pytorch/opacus/2047/workflows/09923631-5734-443b-b639-817f372f8f64/jobs/11263) that only manifest on nightly PyTorch build.

This is a temporary solution - eventually, when enough people have migrated past torch==1.12 we only want to keep current (1.13+) API imtegration

### DIFF
--- a/opacus/grad_sample/gsm_exp_weights.py
+++ b/opacus/grad_sample/gsm_exp_weights.py
@@ -51,5 +51,5 @@ class GradSampleModuleExpandedWeights(AbstractGradSampleModule):
             )(x, *args, **kwargs)
         else:
             return self.call_for_per_sample_grads(
-                self._module, x.shape[0], x, *args, **kwargs
+                module=self._module, batch_size=x.shape[0], args=(x, *args), **kwargs
             )
--- a/opacus/grad_sample/gsm_exp_weights.py
+++ b/opacus/grad_sample/gsm_exp_weights.py
@@ -3,6 +3,9 @@ import torch.nn as nn
 from opacus.grad_sample.gsm_base import AbstractGradSampleModule
 
 
+COMPATIBILITY_API_CUTOFF_VERSION = "1.13.0.dev"
+
+
 class GradSampleModuleExpandedWeights(AbstractGradSampleModule):
     """
     ExpandedWeights-based implementation of AbstractGradSampleModule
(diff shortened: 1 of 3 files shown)


opacus/grad_sample/gsm_exp_weights.py
```diff
@@ -40,6 +43,13 @@ def __init__(
...
-        )
+        if torch.__version__ >= COMPATIBILITY_API_CUTOFF_VERSION:
+            return self.call_for_per_sample_grads(
+                module=self._module,
+                batch_size=x.shape[0],
+                loss_reduction=self.loss_reduction,
+            )(x, *args, **kwargs)
+        else:
+            return self.call_for_per_sample_grads(
+                self._module, x.shape[0], x, *args, **kwargs
```
karthikprasad, line 54:
```suggestion
                module=self._module, batch_size=x.shape[0], x, *args, **kwargs
```

opacus/grad_sample/gsm_exp_weights.py
```diff
@@ -40,6 +43,13 @@ def __init__(
...
 
     def forward(self, x: torch.Tensor, *args, **kwargs):
-        return self.call_for_per_sample_grads(
-            self._module, x.shape[0], x, *args, **kwargs
-        )
+        if torch.__version__ >= COMPATIBILITY_API_CUTOFF_VERSION:
+            return self.call_for_per_sample_grads(
+                module=self._module,
+                batch_size=x.shape[0],
+                loss_reduction=self.loss_reduction,
```
karthikprasad, line 50: Does this mean loss_reduction was not supported before? Should this be called out in our README too?

opacus/privacy_engine.py
```diff
@@ -36,6 +37,13 @@
 from torch.utils.data import DataLoader
 
 
+def _is_ew_compatibility_required(grad_sample_mode: str):
```
karthikprasad, line 40:
```suggestion
def _is_ew_compatible(grad_sample_mode: str):
```

karthikprasad (approved):
accepting since my comments are minor nits

opacus/privacy_engine.py
ffuuugor: Nope
We need `ew_compatibility_mode`, because initial version of ExpandedWeights didn't handle loss reduction properly, so we needed to amend the gradients at the optimizer level.
With the recent version, optimizer is still EW compatible, but doesn't need to know about EW at all (i.e. no `ew_compatibility_mode` is required)

karthikprasad: I see. In that case, two minor suggestions:
1. Add the above explanation as a docstring to the function
2. may be make name more explicit? `_is_ew_compatibility_check_required`?

ffuuugor: yeah, I see how much confusion it creates
I'll follow your suggestions above, but also think we'd better remove `ew_compatibility_mode` completely once torch==1.13 released, dropping support for pre-1.13 `ExpandedWeights`

Issue #454, opened 2022-07-19
Using Opacus with Neural Additive Models 


albertyusun:
## 🐛 Bug

I'm attempting to add differential privacy to [Neural Additive Models (NAMs)](https://arxiv.org/pdf/2004.13912.pdf
) as implemented [here](https://github.com/facebookresearch/nbm-spam/blob/main/nbm_spam/models/concept_nam.py) for interpretable machine learning by Filip Radenovic et al. 

The model, as implemented in Filip's code base ([ConceptNAMNary](https://github.com/facebookresearch/nbm-spam/blob/main/nbm_spam/models/concept_nam.py)), essentially implements a neural net with sequential layers comprised of an arbitrary combination of Conv1d layers, BatchNorm layers, and ReLU activation functions. I used `nam = ModuleValidator.fix(nam)` to fix initial errors about changing Batch Norm to Group Norm. 

When I fit a NAM after using the privacy engine, I receive an error where Opacus doesn't support ConceptNAMNary.  ConceptNAMNary inherits from nn.Module and implements ConceptNNNary, another inheritor of nn.Module. Both ConceptNAMNary and ConceptNNary can be found in the same [file](https://github.com/facebookresearch/nbm-spam/blob/main/nbm_spam/models/concept_nam.py). 

How do I fix the following error? My suspicion is that I have to use Opacus to recognize the layers of the sequential model within the broader wrapper of the model. Do I have to register the ConceptNAMNary "wrapper"? Do I have to 'peel back' ConceptNAMNary and add the privacy engine to a primitive NN with only the Conv1d, GroupNorm, ReLU layers? Any tips on this would be extremely helpful!

## To Reproduce

1. Here is a [Colab Link](https://colab.research.google.com/drive/10bHdIUS6QDwT00Ge54-sIQYbA6Hv-3MR?usp=sharing) to reproduce a simple example. 

2. Initialize a [ConceptNAMNary](https://github.com/facebookresearch/nbm-spam/blob/main/nbm_spam/models/concept_nam.py).

3. Attempt to fit a ConceptNAMNary using the Opacus privacy engine.

```
UnsupportedModuleError: 

[NotImplementedError("Model contains a trainable layer that Opacus doesn't currently support(:

ConceptNAMNary(
(concept_dropout): Dropout(p=0.0, inplace=False)
  (concept_nary_nns): ModuleDict(
    (ord1_net0): ConceptNNNary(
      (model): Sequential(
[...]

PR #458, opened 2022-07-29
More goodness from ExpandedWeights

ffuuugor:
This PR brings few assorted updates related to the new ExpandedWeights API

## Validation
Previously, `ModelValidator` was the one-stop shop for checking the input model: it checked whether we support it (via a call to `GradSampleModule.validate`) and whether the module is DP-valid.

Now with multiple implementations of `GradSampleModule` with varying module support criteria, I propose to separate the responsibilities in the following way:
- `ModuleValidator` is responsible for checking **known** DP violations
- `GradSampleModule` is responsible for checking if it's able to handle given input module. It'll do so to the best of it's abilities: hook-based GSM will perform check in the constructor, while ExpandedWeights-based GSM will do so dynamically during the forward call
- To reflect changes above, we'd need to introduce new tests. I suggest we keep most of the `ModuleValidator` tests + introduce a new one to check the output on a level of `PrivacyEngine.make_private()`

See also discussion in #454 for more context

## Enable some `ExpandedWeights` tests back
We've acknowledged some limitations of `ExpandedWeights` by disabling certain tests. With significant progress being made ([loss reduction fix](https://github.com/pytorch/pytorch/pull/80892), [conv3d fix](https://github.com/pytorch/pytorch/pull/80943), [LayerNorm fix](https://github.com/pytorch/pytorch/pull/80895), [InstanceNorm fix](https://github.com/pytorch/pytorch/pull/79800)) we can enable some of them back. I've also removed some obsolete entries from the Readme

## Scrape EW compatibility API
The initial idea was to temporarily provide support for both iterations of ExpandedWeights: old API in PyTorch 1.12 and new API in 1.13. However, with a significant amount of bugfixes (see paragraph above), it's no longer practical to keep supporting 1.12 - that's a lot of convoluted code for a scant amount of benefits. With that, I've removed all of the compatibility API bits and only allow `ExpandedWeights` usage with the 1.13+

### DIFF
--- a/opacus/tests/privacy_engine_validation_test.py
+++ b/opacus/tests/privacy_engine_validation_test.py
@@ -27,8 +27,8 @@ class BasicSupportedModule(nn.Module):
 class CustomLinearModule(nn.Module):
     def __init__(self, in_features, out_features):
         super().__init__()
-        self._weight = nn.Parameter(torch.empty(out_features, in_features))
-        self._bias = nn.Parameter(torch.empty(out_features))
+        self._weight = nn.Parameter(torch.randn(out_features, in_features))
+        self._bias = nn.Parameter(torch.randn(out_features))
 
     def forward(self, x):
         return F.linear(x, self._weight, self._bias)
@@ -37,17 +37,17 @@ class CustomLinearModule(nn.Module):
 class MatmulModule(nn.Module):
     def __init__(self, input_features, output_features):
         super().__init__()
-        self.weight = nn.Parameter(torch.empty(input_features, output_features))
+        self.weight = nn.Parameter(torch.randn(input_features, output_features))
 
     def forward(self, x):
         return torch.matmul(x, self.weight)
 
 
 class LinearWithExtraParam(nn.Module):
-    def __init__(self):
+    def __init__(self, in_features, out_features):
         super().__init__()
-        self.fc = nn.Linear(5, 8)
-        self.extra_param = nn.Parameter(torch.empty(8, 2))
+        self.fc = nn.Linear(in_features, out_features)
+        self.extra_param = nn.Parameter(torch.randn(out_features, 2))
 
     def forward(self, x):
         x = self.fc(x)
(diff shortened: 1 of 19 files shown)


opacus/validators/module_validator.py
```diff
@@ -59,9 +58,7 @@ def validate(
             errors.append(
                 IllegalModuleConfigurationError("Model needs to be in training mode")
             )
-        # 2. validate that all trainable modules are supported by GradSampleModule.
-        errors.extend(GradSampleModule.validate(module=module, strict=False))
-        # 3. perform module specific validations for trainable modules.
+        # 2. perform module specific validations for trainable modules.
```
karthikprasad, line 61: Good catch!

opacus/tests/privacy_engine_validation_test.py
```diff
@@ -0,0 +1,216 @@
...
+            module=module,
+            optimizer=optim,
+            data_loader=dl,
+            noise_multiplier=1.0,
+            max_grad_norm=1.0,
+            grad_sample_mode="ew",
+        )
+
+        for x in dl:
+            module(x)
```
karthikprasad, line 104: nit: how about just one test with 
```
@given(
  grad_sample_mode=st.sampled_from(["ew","hooks"]),
)
```
?
Not a blocker for this PR at all though.

karthikprasad (approved):
Thank you!
giphy

karthikprasad (changes requested):
Actually, the CI failure looks relevant.

karthikprasad (approved):

opacus/tests/privacy_engine_validation_test.py
ffuuugor: Oh, yes, sorry forgot to respond here.

To be honest, I'm not a big fan of using `hypothesis` for tests where we need to test two branches. Hypothesis samples the parameters randomly. In combination with a limited number of runs, it doesn't actually guarantee that any particular set of parameters will be executed every time you run tests. 
In reality this problem would only surface if the function has a lot of independent parameters, but nevertheless it doesn't look like a good practice to me.

I'd say it'l ok to use to to sample things like dimensions, but if we want to test different flags or implementations, I'd rather use 'unittest.subTest'

PR #465, opened 2022-08-09
fix minimum torch version needed for ExpandedWeights in code

karthikprasad:
Differential Revision: D38554495


opacus/grad_sample/gsm_exp_weights.py
```diff
@@ -32,7 +47,7 @@ def __init__(
             raise ImportError(
                 f"Requested grad_sample_mode=ew, "
                 f"but found PyTorch version={torch.__version__}. "
-                f"ExpandedWeights available for torch>=1.12. "
```
ffuuugor, line 35: good catch!

ffuuugor (approved):
Good one!

PR #471, opened 2022-08-16
reduce logging severity for set_to_none

ffuuugor:
We want to warn users of an unexpected behaviour with `set_to_none` flag. Normally, both `nn.Module` and `Optimizer` let clients choose whether they want to remove the `.grad` attribute altogether or just set it to None.
We, on the other hand, don't want to remove the attributes - it's more convenient to assume the `.grad_sample` attribute is always present. It's not an absolute requirement, but we did that historically and I don't see a case for changing it now.

However, default value for `set_to_none` is False, meaning most users are getting annoying logging notifications on every training step.

### DIFF
--- a/opacus/grad_sample/gsm_base.py
+++ b/opacus/grad_sample/gsm_base.py
@@ -94,7 +94,7 @@ class AbstractGradSampleModule(nn.Module, ABC):
             affects regular gradients. Per sample gradients are always set to None)
         """
         if set_to_none is False:
-            logger.info(
+            logger.debug(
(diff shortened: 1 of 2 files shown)


alexandresablayrolles (approved):
Much needed, thanks! I ended up manually adding set_to_none=False in most of my code...
QQ: is there a way to make this message appear only once?

ffuuugor:
> QQ: is there a way to make this message appear only once?

I mean, we can always set a flag for this on a module/optimizer level - do you think it's worth it?

PR #473, opened 2022-08-17
Include batch_size as an argument, instead of sample_rate

ashkan-software:
Summary:
Other examples in Opacus follow this logic for arguments related to batch size:
- batch_size and batch_size_test

CIFAR-10 still uses the `sample_rate * len(train_data)`.

This diff updates that, to aslo make it clearer (someone was confused [here](https://discuss.pytorch.org/t/the-opacus-example-train-batch-size-vs-sampling-rate/158771))

Differential Revision: D38799564

### DIFF
--- a/examples/cifar10.py
+++ b/examples/cifar10.py
@@ -262,7 +262,7 @@ def main():
 
     train_loader = torch.utils.data.DataLoader(
         train_dataset,
-        batch_size=int(args.sample_rate * len(train_dataset)),
+        batch_size=args.batch_size,
         generator=generator,
         num_workers=args.workers,
         pin_memory=True,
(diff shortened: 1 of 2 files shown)


ffuuugor (approved):
👍

examples/cifar10.py
```diff
@@ -421,11 +421,11 @@ def parse_args():
...
     parser.add_argument(
-        "--sample-rate",
-        default=0.04,
+        "--batch-size",
+        default=2000,
         type=float,
-        metavar="SR",
-        help="sample rate used for batch construction (default: 0.005)",
+        metavar="N",
+        help="train bacth size",
```
karthikprasad, line 428: may be change this to approximate batch size in the help string, since the batch size will not necessarily be 2K.

Issue #432, opened 2022-05-24
Print a warning/throw an error when params in optimizer are not the same as that of module's in `make_private`


karthikprasad:
...code is failing because `model = ModuleValidator.fix(model)` is called after passing `model.parameters()` to optimizer. 
`ModuleValidator.fix(model)` changes some layers of the model (specifically replaces `BatchNorm` with `GroupNorm`) and the parameters of these new layers are not tracked by optimizer.

Calling `fix()` before defining optimizer should work.

That said, your issue does create two action items for us:
1. At the very least, print a warning/throw an error when params in optimizer are not the same as that of modules.
2. Print a more helpful message in case of failures. (not sure how)

_Originally posted by @karthikprasad in https://github.com/pytorch/opacus/issues/431#issuecomment-1136349814_

PR #489, opened 2022-08-31
Better validator for matching parameters

### LINKED ISSUE #477: Model finetuning
## 🚀 Feature
It would be nice to be able to only fine-tune a subset of the model parameters with Opacus. Currently this is not possible since there is a check that module parameters = optimizer parameters (https://github.com/pytorch/opacus/blob/5169ad2a892b5b49373f6727c723f66b510e5a3b/opacus/privacy_engine.py#L385)

ffuuugor:
This PR resolves #477 and addresses various issues regarding parameter validation

## Background

As identified in #432, we want to warn the user if optimizer parameters are different from model parameters (e.g. because the module has been replaced by the ModuleValidator). Optimizer expects the weights to have `.grad_sample` attribute populated by the `GradSampleModule`, so there's no sense in having optimizer with different set of params to the input model

### Issue 1: Comparing values

One problem with how we do it currently is that we're looking for the wrong thing. We use `torch.eq`, which performs elementwise comparison and will output True for identical tensors. The problem is - we don't need identical, we need the *exact same object* - so that optimizer can read per-sample gradients computed by the GradSampleModule.

### Issue 2: Frozen layers

When freezing some model layers, people are free to choose how to initialize their optimizer: with all `model.parameters()` or only  those which are not frozen - it doesn't have any material difference on the training process.
However, if they do the latter, opacus will complain that model and optimizer weights are mismatched.

### This PR

In this PR we address both issues by:
1) Checking if the optimizer parameters is a *subset* of model parameters (not including some model params in the optimizer is perfectly fine)
2) using `in` operator, which checks object and not its value
3) Adding extra tests to cover issues described above

### DIFF
--- a/opacus/tests/privacy_engine_test.py
+++ b/opacus/tests/privacy_engine_test.py
@@ -15,6 +15,7 @@
 
 import abc
 import io
+import itertools
 import math
 import unittest
 from abc import ABC
@@ -81,9 +82,18 @@ class BasePrivacyEngineTest(ABC):
     def _init_vanilla_training(
         self,
         state_dict: Optional[OrderedDict[str, torch.Tensor]] = None,
+        opt_exclude_frozen=False,
     ):
         model = self._init_model()
-        optimizer = torch.optim.SGD(model.parameters(), lr=self.LR, momentum=0)
+        optimizer = torch.optim.SGD(
+            [
+                p
+                for p in model.parameters()
+                if p.requires_grad or (not opt_exclude_frozen)
+            ],
+            lr=self.LR,
+            momentum=0,
+        )
         if state_dict:
             model.load_state_dict(state_dict)
         dl = self._init_data()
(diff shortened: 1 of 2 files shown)


opacus/tests/privacy_engine_test.py
```diff
@@ -98,10 +108,19 @@ def _init_private_training(
         poisson_sampling: bool = True,
         clipping: str = "flat",
         grad_sample_mode="hooks",
+        opt_exclude_frozen=False,
     ):
         model = self._init_model()
         model = PrivacyEngine.get_compatible_module(model)
-        optimizer = torch.optim.SGD(model.parameters(), lr=self.LR, momentum=0)
+        optimizer = torch.optim.SGD(
+            [
```
romovpa, line 116: I would suggest something like this for readability
```python
model.parameters() 
if not opt_exclude_frozen 
else [p for p in model.parameters() if p.requires_grad]
```

Same in the `_init_vanilla_training `

romovpa (approved):
Looks good to me.
Thank you for the detailed description!

karthikprasad (approved):
Thanks for the clean and quick fix!

PR #492, opened 2022-09-01
[functorch] No-Op GradSampleModule

alexandresablayrolles:
TL;DR: Adding a No-Op GradSampleModule in case the grad samples are computed by functorch. The CIFAR10 example has been updated to show a typical use-case for that.

The neat thing about functorch is that it directly gives the per-sample gradients with a couple of lines of code. These per-sample gradients are then manually given to `p.grad_sample` by the end-user.

### DIFF
--- a/opacus/grad_sample/gsm_no_op.py
+++ b/opacus/grad_sample/gsm_no_op.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python3
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


ffuuugor (approved):

opacus/grad_sample/gsm_no_op.py
```diff
@@ -0,0 +1,46 @@
...
+from opacus.grad_sample.gsm_base import AbstractGradSampleModule
+
+
+class GradSampleModuleNoOp(AbstractGradSampleModule):
+    """
+    ExpandedWeights-based implementation of AbstractGradSampleModule
+
+    Computes per-sample gradients using PyTorch built-in mechanism of ExpandedWeights.
+    See README.md for more details
+    """
```
ffuuugor, line 27: Is it relevant?

PR #496, opened 2022-09-05
Fix CIFAR-10 example: --batch-size must be int

romovpa:
Fixing tests:
- a bug introduced in #473
- lint


ffuuugor (approved):
Approving, unit tests are failing in main too

PR #497, opened 2022-09-08
Functorch added to grad sample README

ffuuugor:
As we've landed functorch-backed GradSampleModule, we also want to update the README which helps people navigate different grad samplers

For the content of this readme I've also run benchmarks for all the options - some results are surprising and hard to interpret, but we have mostly consistent picture

## tl;dr
* There's no difference on CPU
* functorch performance depends on the exact GPU setup: same benchmarks show could be up to 4x slower or 2x faster than the baseline depending on the GPU
* EW are consistently 25-30% faster for linear, but not conv

## benchmarks
| device | benchmark | hooks | functorch | ExpandedWeights
|:-------:|:-------:|:-------:|:-------:|:-------:|
| cpu | nn.Conv2d | 1x | 0.9x | 1x |
| cpu | nn.Linear | 1x | 1x | 0.9x |
| cpu | full epoch on CIFAR10 example | 1x | 1.5x | 1x |
| Tesla T4 (Google Colab) | nn.Conv2d | 1x | 4x | 0.9x |
| Tesla T4 (Google Colab) | nn.Linear | 1x | 1.25x | 0.75x |
| A100 (AWS) | nn.Conv2d | 1x | 0.5x | 1x |
| A100 (AWS) | nn.Linear | 1x | 1.5x | 0.75x |
| A100 (AWS) | full epoch on CIFAR10 example | 1x | 1.1x | 0.75x |

FYI @samdow

### DIFF
--- a/examples/char-lstm_README.md
+++ b/examples/char-lstm_README.md
@@ -9,7 +9,7 @@ Download the training zip from https://download.pytorch.org/tutorial/data.zip an
 Run with dp:
 
 ```
-python char-lstm-classification.py --epochs=50 --learning-rate=2.0 --hidden-size=128 --delta=8e-5 --sample-rate=0.05 --n-lstm-layers=1 --sigma=1.0 --max-per-sample-grad-norm=1.5 --device=cuda:0 --data-root="/my/folder/data/names/" --test-every 5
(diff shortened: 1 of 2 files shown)


karthikprasad (approved):
giphy

PR #498, opened 2022-09-08
disable padding=same for experimental hook

ffuuugor:
The tests are failing because our experimental `convolution as a backward` GSM can't handle `padding="same"`

Since this grad sampler is not used by default and mostly exists for aesthetic reasons, I think disabling the test is the best course of action.

### DIFF
--- a/opacus/tests/grad_samples/conv2d_test.py
+++ b/opacus/tests/grad_samples/conv2d_test.py
@@ -24,7 +24,7 @@ from opacus.grad_sample.grad_sample_module import GradSampleModule
 from opacus.utils.tensor_utils import unfold2d
 from torch.testing import assert_allclose
 
-from .common import expander, GradSampleHooks_test, shrinker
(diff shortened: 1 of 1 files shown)


opacus/tests/grad_samples/conv2d_test.py
```diff
@@ -85,20 +85,22 @@ def test_conv2d(
...
-            x,
-            conv,
-            batch_first=True,
-            atol=10e-5,
-            rtol=10e-4,
-            ew_compatible=is_ew_compatible,
-        )
-        GradSampleModule.GRAD_SAMPLERS[nn.Conv2d] = conv2d_gsm
+
+        if padding != "same":
```
karthikprasad, line 89: Can you add a comment here explaining why this check is needed? Essentially copy-pasta the PR description/summary here (and may be also in the actual grad_sampler file)

karthikprasad (approved):
Left one non-blocking comment. Thanks for the quick fix :)

PR #503, opened 2022-09-12
Fix benchmarks

alexandresablayrolles:
Fix micro benchmarks failing in CircleCI.


ffuuugor (approved):
My bad, thanks for fixing

PR #508, opened 2022-09-14
restore benchmarking for dp layers and fix nightly tests

ffuuugor:
Ok, this time it should work.

See successful CircleCI run [here](https://github.com/pytorch/opacus/pull/508/checks?check_run_id=8350960977). Normally we don't run benchmarks for every commit to save time. So here I temporarily added benchmarks to commit tests, made sure they pass and then disabled them back. CircleCI link above include previously failing microbenchmarks

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -247,16 +247,16 @@ commands:
         default: "cpu"
         type: string
       layers:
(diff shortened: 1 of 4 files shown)


alexandresablayrolles (approved):
LGTM

PR #509, opened 2022-09-15
Align make_private_with_epsilon with make_private

### LINKED ISSUE #467: mismatch between the documentation and the behaviour of `make_private_with_epsilon`
`PrivacyEngine.make_private_with_epsilon` does  not behave as explained by the documentation. In particular, `poisson_sampling` can not passed as an argument to this method. I think that the error is due to calling `get_noise_multiplier` (in accountants/utils.py) with `**kwargs`, but `RDPAccountant.get_epsilon` (in `rdp.py`) and  `GaussianAccountant.get_epsilon` (in `gdp.py`) only accepts a fixed number of arguments. Also, I think that the signature of GaussianAccountant.get_epsilon should be `get_epsilon(self, delta: float, poisson_sampling: bool = True) -> float`

alexandresablayrolles:
Fixes #467

### DIFF
--- a/opacus/privacy_engine.py
+++ b/opacus/privacy_engine.py
@@ -524,7 +524,6 @@ class PrivacyEngine:
             loss_reduction=loss_reduction,
             noise_generator=noise_generator,
             grad_sample_mode=grad_sample_mode,
-            max_grad_norm=max_grad_norm,
             poisson_sampling=poisson_sampling,
             clipping=clipping,
         )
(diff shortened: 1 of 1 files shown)


ffuuugor (approved):
Great stuff, thanks!

Accepting to unblock - please make sure tests pass before landing, you've put "max_grad_norm" in there twice

karthikprasad (approved):

PR #519, opened 2022-10-06
Support nn.EmbeddingBag

alexandresablayrolles:
WIP. Passing activations as a list instead of as a tensor.

### DIFF
--- a/opacus/grad_sample/embedding.py
+++ b/opacus/grad_sample/embedding.py
@@ -54,3 +54,25 @@ def compute_embedding_grad_sample(
         torch.backends.cudnn.deterministic = saved
         ret[layer.weight] = grad_sample
     return ret
+
+
+@register_grad_sampler(nn.EmbeddingBag)
+def compute_embeddingbag_gradsampler(layer, inputs, backprops):
+    index, offset = inputs
+    batch_size = offset.shape[0]
+    gsm = torch.zeros(batch_size, layer.num_embeddings, layer.embedding_dim)
+
+    for i in range(batch_size):
+        begin = offset[i]
+        if i < batch_size - 1:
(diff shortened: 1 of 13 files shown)


opacus/grad_sample/embedding.py
```diff
@@ -53,3 +54,25 @@ def compute_embedding_grad_sample(
         torch.backends.cudnn.deterministic = saved
         ret[layer.weight] = grad_sample
     return ret
+
+
+@register_grad_sampler(nn.EmbeddingBag)
+def compute_embeddingbag_gradsampler(layer, inputs, backprops):
```
ffuuugor, line 60: Can we maybe have a test for it, like we have for other supported layers?

ffuuugor (changes requested):
Cool, definitely make sense given that the inputs to the backward hook itself is a list of tensors.
Left one inline comment. Additionally, please take a look at the linter output. Unit test error is irrelevant as it's inherited from main, but the flake is new I think

ffuuugor (approved):

opacus/grad_sample/grad_sample_module.py
```diff
@@ -282,7 +282,7 @@ def capture_activations_hook(
 
         if not hasattr(module, "activations"):
             module.activations = []
-        module.activations.append(forward_input[0].detach())  # pyre-ignore
+        module.activations.append([t.detach() for t in forward_input])  # pyre-ignore
```
karthikprasad, line 285: should this be append or extend?

ffuuugor: I think append is right, as activations is a list of tuples in this case

ffuuugor:
LGTM

PR #528, opened 2022-10-24
Make DPMultiheadAttention drop-in compatible with nn.MultiheadAttention

### LINKED ISSUE #123: DPMultiheadAttention is not drop-in compatible with nn.MultiheadAttention

Wei-1:
## Types of changes

This PR is target to resolve #123 by having an additional re-naming mechanism to match the `state_dict` structure of `nn.MultiheadAttention`.

Run `pytest` within the `opacus` folder:

```
============= 162 passed, 41 skipped, 4411 warnings in 243.18s (0:04:03) =============
```

## Checklist

### DIFF
--- a/opacus/layers/dp_multihead_attention.py
+++ b/opacus/layers/dp_multihead_attention.py
@@ -20,6 +20,8 @@ import torch.nn as nn
 import torch.nn.functional as F
 from torch.nn.parameter import Parameter
 
+from collections import OrderedDict
+
 
 class SequenceBias(nn.Module):
     r"""
@@ -88,11 +90,14 @@ class DPMultiheadAttention(nn.Module):
         add_zero_attn=False,
         kdim=None,
         vdim=None,
+        device=None,
+        dtype=None
     ):
         super(DPMultiheadAttention, self).__init__()
         self.embed_dim = embed_dim
         self.kdim = kdim if kdim is not None else embed_dim
         self.vdim = vdim if vdim is not None else embed_dim
+        self._qkv_same_embed_dim = self.kdim == embed_dim and self.vdim == embed_dim
 
         self.num_heads = num_heads
         self.dropout = dropout
@@ -107,7 +112,7 @@ class DPMultiheadAttention(nn.Module):
 
         # torch.nn.MultiHeadAttention out_proj is _LinearWithBias
         # explicilty setting bias=True for consistent mimicry
-        self.out_proj = nn.Linear(embed_dim, embed_dim, bias=True)
+        self.out_proj = nn.Linear(embed_dim, embed_dim, bias=bias)
 
         self.add_bias_kv = add_bias_kv
         if self.add_bias_kv:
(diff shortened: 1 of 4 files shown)


Wei-1:
Remove for duplicated PR

Issue #522, opened 2022-10-14
The `BatchSplittingSampler` cannot handle empty batches


xichens:
## 🐛 Bug

When poisson sampling is used, empty batches can occur. However, the `BatchSplittingSampler` from `batch_memory_manager.py`, which is called when using the `BatchMemoryManager`, cannot handle empty batches and will throw an error.

## To Reproduce
To reproduce it, see this [colab link](https://colab.research.google.com/drive/1OoP6iIeEkgkOTBGxVued7WSY7k1DEeHn?usp=sharing).

## Expected behavior
The wrapped batch sampler should handle empty batches properly

## Additional context
I think the issue is with [this line](https://github.com/pytorch/opacus/blob/a6c2567491f5089cfc9268a90608992acd4ede5d/opacus/utils/batch_memory_manager.py#L56)
When calling
```
            split_idxs = np.array_split(
                batch_idxs, math.ceil(len(batch_idxs) / self.max_batch_size)
            )
```
the `batch_idxs` can be an empty list since it is from a `UniformWithReplacementSampler`, but `np.array_split` does not expect the first arg to be empty.

xichens:
A follow up issue related to this one. Even without calling the `wrap_data_loader` from `BatchMemoryManager`, the `DPDataLoader` has some issue handling empty batches in certain cases. Currently the `wrapt_collate_with_empty` function [here](https://github.com/pytorch/opacus/blob/a6c2567491f5089cfc9268a90608992acd4ede5d/opacus/data_loader.py#L31) creates empty tensors for empty batches. However, it only sets the shape but not the `dtype` for the empty tensors. By default, these empty tensors will be of `float`. But some modules expect a particular data type as input, for example, for the `torch.nn.Embedding`, the inputs should be of int or Long.

I think the `wrapt_collate_with_empty` should consider the dtypes as well as shapes of the actual sample when creating the empty tensors.

PR #530, opened 2022-10-25
Support empty batches

ffuuugor:
## Background

Poisson sampling can sometimes result in an empty input batch, especially if a sampling rate (i.e. expected batch size) is small. This is not out of the ordinary and should be handled accordingly - gradients (signal) should be set to 0 and noise should still be added.

We've made an [attempt](https://github.com/pytorch/opacus/blob/main/opacus/data_loader.py#L31) to support this behaviour, but it wasn't fully covered with tests and got broken over time. As a result, at the moment we have a DataLoader that is capable of producing zero-sized batches, GradSampleModule that only partially supports them and DPOptimizer that doesn't support them at all

This PR addresses Issue #522 (thanks @xichens for reporting)

## Improvements

This diff fixes the following

* DPOptimizer can now handle empty batches
* BatchMemoryManager can now handle empty batches 
* Adds a PrivacyEngine test with empty batches
* Adds BatchMemoryManager test with empty batches
* DataLoader now respects dtype of the inputs (i.e. empty batches only used to work with float input tensors)
* ExpandedWeights still can's process empty batches, which we call out in our readme (FYI @samdow )


opacus/data_loader.py
```diff
@@ -67,6 +72,19 @@ def shape_safe(x: Any):
...
 
+def dtype_safe(x: Any):
+    """
+    Exception-safe getter for ``dtype`` attribute
+
+    Args:
+        x: any object
+
+    Returns:
+        ``x.shape`` if attribute exists, empty tuple otherwise
```
alexandresablayrolles, line 83: Wrong docstring?

opacus/data_loader.py
```diff
@@ -144,6 +162,7 @@ def __init__(
                 generator=generator,
             )
         sample_empty_shapes = [[0, *shape_safe(x)] for x in dataset[0]]
+        dtypes = [dtype_safe(x) for x in dataset[0]]
```
alexandresablayrolles, line 165: If this is consumed by  wrap_collate_with_empty, we need the dtype to be an actual `dtype` and not a `type` right?

opacus/grad_sample/conv.py
```diff
@@ -41,6 +41,14 @@ def compute_conv_grad_sample(
         backprops: Backpropagations
     """
     n = activations.shape[0]
+    if n == 0:
+        # Empty batch
+        ret = {}
+        ret[layer.weight] = torch.zeros_like(layer.weight).unsqueeze(0)
+        if layer.bias is not None and layer.bias.requires_grad:
+            ret[layer.bias] = torch.zeros_like(layer.bias).unsqueeze(0)
+        return ret
```
alexandresablayrolles, line 50: Why does conv need a special treatment? What happens with other layers?

alexandresablayrolles (commented):
Looks good overall but left a few comments.

opacus/data_loader.py
ffuuugor: True

opacus/data_loader.py
ffuuugor: torch.zeros support normal python types (int/float/bool) just fine:
```
> torch.zeros((2,2), dtype=int)
tensor([[0, 0],
        [0, 0]])

> torch.zeros((2,2), dtype=float)  
tensor([[0., 0.],
        [0., 0.]], dtype=torch.float64)

> torch.zeros((2,2), dtype=bool)   
tensor([[False, False],
        [False, False]])
```

opacus/grad_sample/conv.py
ffuuugor: That's down to specific grad sampler implementation.
Most grad samplers we have rely on einsums, which are generally pretty good with handling 0-sized vectors.
With conv in particular the culprit is the following line: 

```
backprops = backprops.reshape(n, -1, activations.shape[-1])
```

That said, it is a good point that we want to be sure all of the layers can handle it - which current tests only partially do. We have a PrivacyEngine test for an empty batch, but nothing on the grad sampler level - I'll check if there's an easy way to do it and update the PR

opacus/data_loader.py
joserapa98: Hi there! I've been following the discussion since I ran into the empty batches problem when using poission sampling with small sampling rate. When I tested this implementation with my own project, it didn't work for me, since it does not support more complex labels. If labels are just numbers (`int`, `float`, `bool`), it's ok, but if you have maybe a label formed by a tuple of numbers, its type will be `tuple`, thus causing an error in `torch.zeros`. 

I don't know if you expect this kind of things to be supported, but as they work in standard PyTorch (and in fact in Opacus when batches are not empty), maybe it is something worth to be aware of.

ffuuugor: That's an interesting point, thanks

Just thinking out loud, what would be a good way to support this? In case the original label is a tuple, how does collate function handles it - would it output multiple tensors per label?

joserapa98: If you have a dataset in which each sample is of type, say, `tuple(torch.Tensor, tuple(int, int))`, the dataloader would return `tuple(torch.Tensor, tuple(torch.Tensor, torch.Tensor))`, where now each tensor has an extra dimension for the batch. Something similar would happen if labels are given as lists, dicts, etc.

This is the code snippet that manages this cases:

    if isinstance(elem, collections.abc.Mapping):
        try:
            return elem_type({key: collate([d[key] for d in batch], collate_fn_map=collate_fn_map) for key in elem})
        except TypeError:
            # The mapping type may not support `__init__(iterable)`.
            return {key: collate([d[key] for d in batch], collate_fn_map=collate_fn_map) for key in elem}
    elif isinstance(elem, tuple) and hasattr(elem, '_fields'):  # namedtuple
        return elem_type(*(collate(samples, collate_fn_map=collate_fn_map) for samples in zip(*batch)))
    elif isinstance(elem, collections.abc.Sequence):
        # check to make sure that the elements in batch have consistent size
        it = iter(batch)
        elem_size = len(next(it))
        if not all(len(elem) == elem_size for elem in it):
            raise RuntimeError('each element in list of batch should be of equal size')
        transposed = list(zip(*batch))  # It may be accessed twice, so we use a list.


        if isinstance(elem, tuple):
            return [collate(samples, collate_fn_map=collate_fn_map) for samples in transposed]  # Backwards compatibility.
        else:
            try:
                return elem_type([collate(samples, collate_fn_map=collate_fn_map) for samples in transposed])
            except TypeError:
                # The sequence type may not support `__init__(iterable)` (e.g., `range`).
                return [collate(samples, collate_fn_map=collate_fn_map) for samples in transposed]


https://github.com/pytorch/pytorch/blob/8349bf1cd1d5df7be73b194940bcf96209159f40/torch/utils/data/_utils/collate.py#L126-L149

I guess a proper solution for supporting empty batches would be to recycle this code but returning tuples/dicts/lists/... of empty tensors with `torch.zeros`, so that the types are still preserved, though filled with empty tensors.

opacus/data_loader.py
```diff
@@ -29,7 +29,9 @@
 
 
 def wrap_collate_with_empty(
-    collate_fn: Optional[_collate_fn_t], sample_empty_shapes: Sequence
+    collate_fn: Optional[_collate_fn_t],
+    sample_empty_shapes: Sequence[torch.Size],
```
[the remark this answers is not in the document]
ffuuugor: Good question - I'm not aware of what the best practices are.
List os probably out of the question - it's mutable and can never be returned from shape-related methods.

torch.Size vs tuple is more interesting. Obv, passing tuple will work (after all, torch.Size is a tuple subclass with few bells and whistles), but torch will always return Size.

Given suggestive nature of typehints, does it make sense to keep torch.Size to indicate it needs to be returned from torch.Tensor.shape?

alexandresablayrolles (approved):
LGTM

Issue #551, opened 2022-12-25
cannot install opacus with torch(+gpu)


giladcohen:
Upon calling
`pip install opacus`
pip is trying to delete my installed torch(GPU) version and install a CPU version instead.

## To Reproduce
> :warning: **We cannot help you without you sharing reproducible code**. Do not ignore this part :)
1. install pytorch 1.13.1:
`pip install torch torchvision torchaudio --extra-index-url https://download.pytorch.org/whl/cu117`
2. install opacus:
`pip install opacus`

This is what I got:
```
Collecting opacus
  Downloading opacus-1.3.0-py3-none-any.whl (216 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 216.9/216.9 kB 1.5 MB/s eta 0:00:00
Collecting functorch
  Using cached functorch-1.13.0-py2.py3-none-any.whl (2.1 kB)
Requirement already satisfied: torch>=1.8 in /home/gilad/venv_py37_new/lib/python3.7/site-packages (from opacus) (1.13.1+cu117)
Requirement already satisfied: scipy>=1.2 in /home/gilad/venv_py37_new/lib/python3.7/site-packages (from opacus) (1.7.3)
Requirement already satisfied: numpy>=1.15 in /home/gilad/venv_py37_new/lib/python3.7/site-packages (from opacus) (1.19.5)
Requirement already satisfied: opt-einsum>=3.3.0 in /home/gilad/venv_py37_new/lib/python3.7/site-packages (from opacus) (3.3.0)
Requirement already satisfied: typing-extensions in /home/gilad/venv_py37_new/lib/python3.7/site-packages (from torch>=1.8->opacus) (4.0.1)
Collecting torch>=1.8
  Downloading torch-1.13.0-cp37-cp37m-manylinux1_x86_64.whl (890.2 MB)
```

It seems that `functorch` is forcing a non cuda pytorch version.
I experienced the same behavior also when I tried to install opacus for torch==1.12.1+cu113.

I am using python vertion 3.7.9

## Environment
```
(venv_py37_new) gilad@Yoshua:~/workspace$ python collect_env.py
Collecting environment information...
PyTorch version: 1.13.1+cu117
Is debug build: False
CUDA used to build PyTorch: 11.7
ROCM used to build PyTorch: N/A

OS: Ubuntu 18.04.5 LTS (x86_64)
[...]

PR #12, opened 2020-04-01
Change in computing RDP

astorfi:
## Change in computing RDP

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Docs change / refactoring / dependency upgrade

## Motivation and Context / Related issue

The computation of RDP is inconsistent with the original paper. A minor change is performed to correct the calculations.
Please refer to [this issue](https://github.com/facebookresearch/pytorch-dp/issues/11) that I created.

## How Has This Been Tested (if it applies)
The code has been run and tested using ``pytest`` and ``pytest-cov``.

## Checklist

- [x] The documentation is up-to-date with the changes I made.
- [x] I have read the **CONTRIBUTING** document and completed the CLA (see **CONTRIBUTING**).
- [x] All tests passed, and additional code has been covered with new tests.

PR #555, opened 2023-01-14
Update PyTorch requirement to >=1.13

ffuuugor:
functorch is included in the core PyTorch [since 1.13](https://github.com/pytorch/functorch/releases/tag/v1.13.0).
They, however continue to maintain separate PyPI package, which in fact causes issues: #551, [conda forge #12](https://github.com/conda-forge/opacus-feedstock/pull/12#issuecomment-1382580092)

I suggest we remove functorch dependency and force `torch>=1.13`. If anyone needs older pytorch compatibility, they're free to use older opacus versions. 

To accompany this change, I've also updated the list of pytorch/python versions we test installation on.

Nice bonus of doing this is we can remove all the awkward "check if functorch available" code.

Bonus content: I've also updated CircleCI badge to only show workflows on the main branch (was surprised to see green badge now, before merging the fix)


karthikprasad (approved):
Love the simplification!
Thanks for this and for finding the conda issue!

One question: does this mean csprng is support is gone for good?

Issue #375, opened 2022-03-11
max_grad_norm schedule


ffuuugor:
`max_grad_norm` is functionally very similar to `noise_multplier` - it's also stored in the optimizer field as a scalar and never updated.

We have convenient scheduling option for optimizer, but not for `max_grad_norm`.
I can't point to any particular paper proposing this, but some (e.g. https://arxiv.org/pdf/2108.01624.pdf) show that over time singal-to-noise ratio can decline due to natural decline in gradient norm (expected in many training scenarions)

PR #556, opened 2023-01-20
Add Clipping schedulers

Darktex:
Summary:
This diff introduces gradient clipping schedulers that can be used to vary gradient clipping throughout training.

Addresses #375 in OSS.

Differential Revision: D42644261

### DIFF
--- a/opacus/tests/privacy_engine_test.py
+++ b/opacus/tests/privacy_engine_test.py
@@ -19,7 +19,7 @@ import itertools
 import math
 import unittest
 from abc import ABC
-from typing import Optional, OrderedDict, Type
+from typing import Optional, OrderedDict
 from unittest.mock import MagicMock, patch
 
 import hypothesis.strategies as st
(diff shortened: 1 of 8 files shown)


opacus/tests/privacy_engine_test.py
```diff
@@ -550,10 +550,13 @@ def test_parameters_match(self):
             )
 
     @given(
-        noise_scheduler=st.sampled_from([None, StepNoise]),
+        has_noise_scheduler=st.booleans(),
+        has_grad_clip_scheduler=st.booleans(),
     )
     @settings(deadline=None)
-    def test_checkpoints(self, noise_scheduler: Optional[Type[StepNoise]]):
+    def test_checkpoints(
```
karthikprasad, line 557: Could you break this function into multiple smaller functions to make flake8 happy?

karthikprasad (approved):
Thanks for taking this one on. The changes look good to me, but be sure to fix the lint before merging. :)

opacus/tests/privacy_engine_test.py
Darktex: Yeah, looks like that function was right at the limit for complexity and I crossed it :D Will refactor

PR #574, opened 2023-03-13
Change z in PRV accountant to remove Userwarning for log(x<0)

alexandresablayrolles:
Change definition of `z = np.log((np.exp(t) + q - 1) / q)` to `z = np.log(np.where(t > np.log(1 - q), (np.exp(t) + q - 1) / q, 1))` to avoid userwarning in PRV.


alexandresablayrolles:
@ffuuugor also, any reason we have 'rdp' as default accoutant in get_noise_multiplier and not 'prv'?

Issue #567, opened 2023-02-21
Same uncertainty value for every input


zedoul:
## 🐛 Bug

Hi,

First of all, thank you for a great open source product! Here I think I found a potential issue on Opacus, so I would like to report the bug.

A short summary: An opacus-based classfication model generates the same uncertainty value for every input, when the number of feature is equal to 4, on Wine dataset.

My hypothesis at this moment is that this error is able to be reproduced when 1) a classification model 2) a tabular data and 3) the number of X is 4. The last condition may sound very strange and I try to reproduce the bug with another conditions, but that's where my conclusion is ended up right now. I am pretty sure that the number of features is not something related to a root cause and there must be another datasets that may cause the same error but without the third condition. However, at least I hope my sample code is something that is reproduciable on your local. Also, the data works fine on TensorFlow privacy equivalent implementation.

Here is my reproduced result. I tried to use Colab as you recommended but I unfortunately do not know how to use it. Instead, I copied and pasted the code and attached dataset. The dataset that I used is [the kaggle Wine data](https://www.kaggle.com/datasets/uciml/red-wine-quality-cortez-et-al-2009?resource=download). But, as I said, I faced the error on another data too.

My sample code is not really great, e.g., you may find some hyperparamters could be further improved, etc. Regardless, I expect that any output from Opacus-enhanced version of PyTorch should produce different uncertainty values if the inputs are different, especially when a normal PyTorch, without Opacus, generates different uncertainty values.

```py
import torch
import numpy as np
import pandas as pd
import torch.nn as nn
import torch.optim as torch_optim
import torch.nn.functional as F

import opacus

batch_size = 50
epochs = 10
lr = 0.0001
n_class = 10

df = pd.read_csv("wine.csv")


X = df.iloc[:, 5:9]
y = df.iloc[:, -1]
X = X.to_numpy()
y = y.to_numpy()


assert(X.shape[1] == 4)

from sklearn.model_selection import train_test_split
from sklearn.preprocessing import LabelEncoder

train_priv_X, test_pub_X, train_priv_Y, test_pub_Y = train_test_split(X, y, test_size=0.2, stratify=y)

train_priv_Y = LabelEncoder().fit_transform(train_priv_Y)
test_pub_Y = LabelEncoder().fit_transform(test_pub_Y)

from torch.utils.data import Dataset, DataLoader
import numpy as np

class PowerDataset(Dataset):
    def __init__(self, X, Y):
        X = X.copy()
[...]

PR #580, opened 2023-03-30
Pass `num_groups` arg to ModelValidator.fix

ffuuugor:
## Problem
As highlighted by #567, end user have little control over how exactly `ModelValidator.fix()` deals with BatchNorms.
For example, our approach to choosing number of groups is `gcd(32, module.num_features)`, which is fine for most cases, but can break a model occasionally (see #567 for a demonstration)

## Solution
Pass `num_groups` as kwarg, allowing clients to control the behaviour

### DIFF
--- a/opacus/validators/batch_norm.py
+++ b/opacus/validators/batch_norm.py
@@ -15,7 +15,7 @@
 
 import logging
 import math
-from typing import List, Union
+from typing import List, Optional, Union
 
 import torch.nn as nn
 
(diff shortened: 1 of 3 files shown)


opacus/validators/batch_norm.py
```diff
@@ -86,9 +93,10 @@ def _batchnorm_to_groupnorm(module: BATCHNORM) -> nn.GroupNorm:
         A default value of 32 is chosen for the number of groups based on the
         paper *Group Normalization* https://arxiv.org/abs/1803.08494
     """
-    return nn.GroupNorm(
-        math.gcd(32, module.num_features), module.num_features, affine=module.affine
-    )
+    if num_groups is None:
+        num_groups = math.gcd(32, module.num_features)
+
+    return nn.GroupNorm(num_groups, module.num_features, affine=module.affine)
```
karthikprasad, line 99: Might want to check that `num_features` is divisible by `num_groups`; and raise an exception otherwise.

opacus/validators/batch_norm.py
```diff
@@ -62,16 +62,23 @@ def fix(module: BATCHNORM, **kwargs) -> Union[nn.GroupNorm, INSTANCENORM]:
...
         _batchnorm_to_instancenorm(module)
         if is_replace_bn_with_in
-        else _batchnorm_to_groupnorm(module)
+        else _batchnorm_to_groupnorm(module, num_groups)
     )
 
 
-def _batchnorm_to_groupnorm(module: BATCHNORM) -> nn.GroupNorm:
+def _batchnorm_to_groupnorm(
+    module: BATCHNORM, num_groups: Optional[int] = None
```
karthikprasad, line 80:
```suggestion
    module: BATCHNORM, *, num_groups: Optional[int] = None
```

opacus/validators/batch_norm.py
ffuuugor: Actually I didn't do any validation on purpose - exception will be thrown from `nn.GroupNorm` constructor immediately and it will be a good one, so no reason to wrap it with our own

karthikprasad: Point taken!

karthikprasad (approved):

Issue #328, opened 2022-01-18
deprecation message for non-full backward hook


Solosneros:
## 🐛 Bug

I am getting the following warning when using opacus on my system:
`/usr/local/lib/python3.7/dist-packages/torch/nn/modules/module.py:1025: UserWarning: Using a non-full backward hook when the forward contains multiple autograd Nodes is deprecated and will be removed in future versions. This hook will be missing some grad_input. Please use register_full_backward_hook to get the documented behavior.
  warnings.warn("Using a non-full backward hook when the forward contains multiple autograd Nodes "`

## To Reproduce
(sorry, I had problems running google colab)..

1. Execute the minimal example code given below:

```
import numpy as np
import opacus
import torch
import torch.nn as nn
from opacus.utils.uniform_sampler import UniformWithReplacementSampler
from torch.utils.data import DataLoader, TensorDataset, Dataset


DEVICE = "cuda:0"

BATCH_SZ = 8
FEATURE_DIM = 16
N_CLASSES = 2
DATASET_LEN = 50_000

class DummyDataset(Dataset):
    def __init__(self):
        pass

    def __getitem__(self, i):
        x = torch.randn([FEATURE_DIM]).to(DEVICE) 
        y = 0
        return x, y

    def __len__(self):
        return DATASET_LEN

train_ds = DummyDataset()
train_loader = DataLoader(train_ds, BATCH_SZ, shuffle=False)

model = nn.Linear(FEATURE_DIM, N_CLASSES).to(DEVICE)
criterion = nn.CrossEntropyLoss()

privacy_engine = opacus.PrivacyEngine()
model, optimizer, train_loader = privacy_engine.make_private(
        module=model,
        optimizer=torch.optim.SGD(model.parameters(), 0.01),
        data_loader=train_loader,
        noise_multiplier=1,
        max_grad_norm=10,
    )

for epoch in range(2):
  for x, y in train_loader:
        x = x.to(DEVICE)
        y = y.to(DEVICE)
        optimizer.zero_grad()
        out = model(x)
        loss = criterion(out, y)
        loss.backward()
        optimizer.step()
```

## Expected behavior

No error message.

## Environment

PyTorch version: 1.10.1
Is debug build: False
CUDA used to build PyTorch: 10.2
ROCM used to build PyTorch: N/A

OS: Ubuntu 20.04.3 LTS (x86_64)
GCC version: (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
[...]

PR #581, opened 2023-04-06
Clean-up unnecessary warnings (including update to PyTorch 2.0)

ffuuugor:
This PR is a collection of smaller fixes that will save us some deprecation issues in the future

## 1. Updating to PyTorch 2.0

**Key files: grad_sample/functorch.py, requirements.txt**

`functorch` has been a part of core PyTorch since 1.13.
Now they're going a step further and changing the API, while deprecating the old one.

There's a [guide](https://pytorch.org/docs/master/func.migrating.html) on how to migrate. TL;DR - `make_functional` will no longer be part of the API, with `torch.func.functional_call()` being (non drop-in) replacement. 

They key difference for us is `make_functional()` creates a fresh copy of the module, while `functional_call()` uses existing module. As a matter of fact, we need the fresh copy (otherwise all the hooks start firing and you enter nested madness), so I've copy-pasted a [gist](https://gist.github.com/zou3519/7769506acc899d83ef1464e28f22e6cf) from the official guide on how to get a full replacement for `make_functional`.

## 2. New mechanism for gradient accumulation detection

**Key file: privacy_engine.py, grad_sample_module.py**

As [reported](https://discuss.pytorch.org/t/gan-raises-userwarning-using-a-non-full-backward-hook-when-the-forward-contains-multiple/175638/2) on the forum, clients are still getting "non-full backward hook" warning even when using `grad_sample_mode="ew"`. Naturally, `functorch` and `hooks` modes rely on backward hooks and can't be migrated to full hooks because [reasons](https://github.com/pytorch/opacus/issues/328#issuecomment-1027361968). However, `ew` doesn't rely on hooks and it's unclear why the message should appear.

The reason, however, is simple. If the client is using poisson sampling we add an extra check to prohibit gradient accumulation (two poisson batches combined is not a poisson batch), and we do that by the means of backward hooks.

~In this case, backward hook serves a simple purpose and there shouldn't be any problems with migrating to the new method, however that involved changing the checking method. That's because `register_backward_hook` is called *after* hooks on submodule, but `register_full_backward_hook` is called before.~

Strikethrough solution didn't work, because hook order execution is weird for complex graphs, e.g. for GANs. For example, if your forward call looks like this:
```
Discriminator(Generator(x))
```
then top-level module hook will precede submodule's hooks for `Generator`, but not for `Discriminator`

As such, I've realised that gradient accumulation is not even supported in `ExpandedWeights`, so we don't have to worry about that. And the other two modes are both hooks-based, so we can just check the accumulation in the existing backward hook, no need for an extra hook. Deleted some code, profit.

## 3. Refactoring `wrap_collate_with_empty` to please pickle

Now here're two facts I didn't know before

1) You can't pickle a nested function, e.g. you can't do the following
```python
def foo():
    def bar():
        <...>
    
    return bar

pickle.dump(foo(), ...)
```

2) Whether or not `multiprocessing` uses pickle is python- and platform- dependant.

This affects our tests when we test `DataLoader` with multiple workers. As such, our data loaders tests:
* Pass on CircleCI with python3.9
* Fail on my local machine with python3.9
* Pass on my local machine with python3.7

I'm not sure how cow common the issue is, but it's safer to just refactor `wrap_collate_with_empty` to avoid nested functions.

## 4. Fix benchmark tests

We don't really run `benchmarks/tests` on a regular basis, and some of them were broken since we've upgraded to PyTorch 1.13 (`API_CUTOFF_VERSION` doesn't exist anymore) 

## 4. Fix flake8 config

Flake8 config no [longer support](https://flake8.pycqa.org/en/latest/user/configuration.html) inline comments, fix is due

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,9 +13,9 @@ commands:
           name: "Setup Python v3.9 environment"
           command: |
             cd /opt/circleci/.pyenv && git pull && cd -
-            pyenv install -s 3.9
-            pyenv global 3.9
-            pyenv local 3.9
+            pyenv install -s 3.9.4
+            pyenv global 3.9.4
+            pyenv local 3.9.4
             pyenv versions
             echo "In venv: $(pyenv local) - $(python -V), $(pip -V)"
             sudo "$(which python)" -m pip install --upgrade pip
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,16 +6,16 @@ version: 2.1
 
 commands:
 
-  py_3_7_setup:
-    description: "Install and switch to Python 3.7.5; also install pip and pytest."
+  py_3_9_setup:
+    description: "Install and switch to Python 3.9; also install pip and pytest."
     steps:
       - run:
-          name: "Setup Python v3.7.5 environment"
+          name: "Setup Python v3.9 environment"
           command: |
             cd /opt/circleci/.pyenv && git pull && cd -
-            pyenv install -s 3.7.5
-            pyenv global 3.7.5
-            pyenv local 3.7.5
+            pyenv install -s 3.9
+            pyenv global 3.9
+            pyenv local 3.9
             pyenv versions
             echo "In venv: $(pyenv local) - $(python -V), $(pip -V)"
             sudo "$(which python)" -m pip install --upgrade pip
@@ -283,23 +283,15 @@ commands:
 
 jobs:
 
-  lint_py37_torch_release:
+  lint_py39_torch_release:
     docker:
-      - image: cimg/python:3.7.5
+      - image: cimg/python:3.9
     steps:
       - checkout
       - pip_dev_install
       - lint_flake8
       - lint_black
       - isort
-
-  unittest_py37_torch_release:
-    docker:
-      - image: cimg/python:3.7.5
-    steps:
-      - checkout
-      - pip_dev_install
-      - unit_tests
 
   unittest_py38_torch_release:
     docker:
@@ -328,10 +320,10 @@ jobs:
 
   prv_accountant_values:
     docker:
-      - image: cimg/python:3.7.5
+      - image: cimg/python:3.9
     steps:
       - checkout
-      - py_3_7_setup
+      - py_3_9_setup
       - pip_dev_install
       - run:
           name: "Unit test prv accountant"
@@ -339,23 +331,23 @@ jobs:
           command: |
             python -m unittest opacus.tests.prv_accountant
 
-  integrationtest_py37_torch_release_cpu:
+  integrationtest_py39_torch_release_cpu:
     docker:
-      - image: cimg/python:3.7.5
+      - image: cimg/python:3.9
     steps:
       - checkout
-      - py_3_7_setup
+      - py_3_9_setup
       - pip_dev_install
       - mnist_integration_test:
           device: "cpu"
 
-  integrationtest_py37_torch_release_cuda:
+  integrationtest_py39_torch_release_cuda:
     machine:
       resource_class: gpu.nvidia.small.multi
       image: ubuntu-2004-cuda-11.4:202110-01
     steps:
       - checkout
-      - py_3_7_setup
+      - py_3_9_setup
       - pip_dev_install
       - run_nvidia_smi
       - mnist_integration_test:
@@ -369,13 +361,13 @@ jobs:
       - dcgan_integration_test:
           device: "cuda"
 
-  micro_benchmarks_py37_torch_release_cuda:
+  micro_benchmarks_py39_torch_release_cuda:
     machine:
       resource_class: gpu.nvidia.small.multi
       image: ubuntu-2004-cuda-11.4:202110-01
     steps:
       - checkout
-      - py_3_7_setup
+      - py_3_9_setup
       - pip_dev_install
       - run_nvidia_smi
       - benchmark_layers_integration_test:
@@ -459,7 +451,7 @@ jobs:
       image: ubuntu-2004-cuda-11.4:202110-01
     steps:
       - checkout
-      - py_3_7_setup
+      - py_3_9_setup
       - pip_dev_install
       - run_nvidia_smi
       - command_unit_tests_multi_gpu
@@ -493,9 +485,7 @@ workflows:
       not:
         equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
     jobs:
-      - lint_py37_torch_release:
-          filters: *exclude_ghpages
-      - unittest_py37_torch_release:
+      - lint_py39_torch_release:
           filters: *exclude_ghpages
       - unittest_py38_torch_release:
           filters: *exclude_ghpages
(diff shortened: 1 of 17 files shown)


.circleci/config.yml
```diff
@@ -6,16 +6,16 @@ version: 2.1
 
 commands:
 
-  py_3_7_setup:
-    description: "Install and switch to Python 3.7.5; also install pip and pytest."
+  py_3_9_setup:
```
karthikprasad, line 9: Why is this change required? Is it necessary for pytorch 2.0?

opacus/data_loader.py
```diff
@@ -28,6 +29,21 @@
 logger = logging.getLogger(__name__)
 
 
+def collate(
+    batch: List[torch.Tensor],
```
karthikprasad, line 33:
```suggestion
def collate(
    *,
    batch: List[torch.Tensor],
```

opacus/data_loader.py
```diff
@@ -28,6 +29,21 @@
...
+    collate_fn: Optional[_collate_fn_t],
+    sample_empty_shapes: Sequence[Tuple],
+    dtypes: Sequence[Union[torch.dtype, Type]],
+):
```
karthikprasad, line 37: nit: Now that this is no longer an inner function, may be some docstrings?

opacus/grad_sample/functorch.py
```diff
@@ -1,5 +1,29 @@
+import copy
+
+import torch
 import torch.nn as nn
 from opacus.layers.dp_rnn import RNNLinear
+from torch.func import grad, vmap
+
+
+# https://gist.github.com/zou3519/7769506acc899d83ef1464e28f22e6cf
+def make_functional(mod, disable_autograd_tracking=False):
```
karthikprasad, line 10: nit: doc string :)

opacus/privacy_engine.py
```diff
@@ -68,18 +68,12 @@ def forbid_accumulation_hook(
         return
 
     for _, p in trainable_parameters(module):
-        if p.grad_sample is not None:
-            if isinstance(p.grad_sample, torch.Tensor):
-                accumulated_iterations = 1
-            elif isinstance(p.grad_sample, list):
-                accumulated_iterations = len(p.grad_sample)
-
-            if accumulated_iterations > 1:
```
karthikprasad, line 77: Is this change a consequence of moving to `register_backward_hook ` ?

opacus/tests/grad_samples/common.py
```diff
@@ -59,10 +59,7 @@ def run_test(
         ):
             grad_sample_modes = ["hooks"]
 
-        if ew_compatible and batch_first and torch.__version__ >= (1, 13):
-            grad_sample_modes += ["ew"]
-
-        for loss_reduction in ["sum", "mean"]:
+        for loss_reduction in ["mean"]:
```
karthikprasad, line 62: Sorry, what happened here?

karthikprasad (commented):
Great PR, and as usual, even greater PR summary! Thanks @ffuuugor 
I do have some questions and nits, but LGTM otherwise. :)

.circleci/config.yml
ffuuugor: yup, torch 2.0 is [not available](https://app.circleci.com/pipelines/github/pytorch/opacus/2700/workflows/60b99641-92fa-44ad-91c5-5ea3279cd271/jobs/15360) for python3.7 and older (at least on PyPI)

opacus/data_loader.py
ffuuugor: good point

opacus/privacy_engine.py
ffuuugor: Indeed. 

`register_backward_hook` hooks are called after submodule-level hooks, while `register_full_backward_hook` is called before we compute per sample gradients for the current iteration

ffuuugor: That said, I'm looking at the failing GAN example and I'm not so sure about this anymore

opacus/grad_sample/functorch.py
```diff
@@ -1,5 +1,40 @@
+import copy
+
+import torch
 import torch.nn as nn
 from opacus.layers.dp_rnn import RNNLinear
+from torch.func import grad, vmap
+
+
+# https://gist.github.com/zou3519/7769506acc899d83ef1464e28f22e6cf
+def make_functional(mod: nn.Module, disable_autograd_tracking: bool = False):
```
alexandresablayrolles, line 10: Should we try to integrate this a bit more to Opacus? Maybe in a later PR?

alexandresablayrolles (approved):
Overall LGTM but some small comments

examples/cifar10.py
```diff
@@ -139,10 +141,9 @@ def train(args, model, train_loader, optimizer, privacy_engine, epoch, device):
     top1_acc = []
 
     if args.grad_sample_mode == "no_op":
-        from functorch import grad_and_value, make_functional, vmap
-
         # Functorch prepare
         fmodel, _fparams = make_functional(model)
+        # params = dict(model.named_parameters())
```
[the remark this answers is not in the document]
ffuuugor: yep, thanks

opacus/grad_sample/functorch.py
ffuuugor: You mean using the native `torch.func.functional_call` in Opacus?

I'm not sure it'll look significantly different from the current solution. The key part of `make_functional` is deep copying the model, which we'll need to do anyway. The rest is very similar to how *native* API would look like

opacus/data_loader.py
karthikprasad: Thanks. Although, the style of docstrings is different from the ususal style.

.circleci/flake8_config.ini
```diff
@@ -5,60 +5,113 @@ max-line-length = 80
...
-  E502, # The backslash is redundant between brackets.
-  E701, # Multiple statements on one line (colon).
-  E702, # Multiple statements on one line (semicolon).
-  E703, # Statement ends with a semicolon.
-  E704, # Multiple statements on one line (def).
-  W291, # Trailing whitespace.
-  W292, # No newline at end of file.
-  W293, # Blank line contains whitespace.
-  W391, # Blank line at end of file.
+  # B950: Line too long. (Use `arc lint`'s LINEWRAP instead)
```
karthikprasad, line 8: Are these changes intentional?

opacus/grad_sample/functorch.py
```diff
@@ -7,14 +7,25 @@
 
 
 # https://gist.github.com/zou3519/7769506acc899d83ef1464e28f22e6cf
-def make_functional(mod, disable_autograd_tracking=False):
+def make_functional(mod: nn.Module, disable_autograd_tracking: bool = False):
+    """
```
karthikprasad, line 11: docstring style is different here too.

opacus/grad_sample/functorch.py
```diff
@@ -7,14 +7,25 @@
...
+    :return:
+    """
     params_dict = dict(mod.named_parameters())
     params_names = params_dict.keys()
     params_values = tuple(params_dict.values())
 
     stateless_mod = copy.deepcopy(mod)
     stateless_mod.to("meta")
 
+    if hasattr(stateless_mod, "allow_grad_accumulation"):
```
karthikprasad, line 26: Wouldn't this be true for all GSMs?

opacus/data_loader.py
ffuuugor: I relied on the IDE too much, thanks for catching!

.circleci/flake8_config.ini
ffuuugor: Yes (I've actually had to learn a new PyCharm shortcut to make it happen).
Flake8 config no [longer support](https://flake8.pycqa.org/en/latest/user/configuration.html) inline comments, so the fix is due

I've updated the PR description too.

PR #589, opened 2023-05-09
Disable memory benchmarking

ffuuugor:
Our tests have been red for a while due to failing memory bechmarks.

## Issue

When benchmarking opacus we run the training script multiple times within one process:

```
for i in range(args.num_runs):
    run_layer_benchmark(
        ...
    )
```

We use built-in pytorch tools to check memory stats. Crucially, we verify that `torch.cuda.memory_allocated()` is 0 before the run starts. Normally, it should be 0, as all previous tensors are out of scope and should have been collected.  

It all worked fine until something changed and some GPU memory stayed allocated between runs. No idea why, but explicit cache clearing or object deletion didn't help.

So I gave up and disabled memory benchmarking, since it seems like it's not a complicated thing to do due to some PyTorch update

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -500,6 +500,8 @@ workflows:
           filters: *exclude_ghpages
       - prv_accountant_values:
           filters: *exclude_ghpages
+      - micro_benchmarks_py39_torch_release_cuda:
+          filters: *exclude_ghpages
 
   nightly:
     when:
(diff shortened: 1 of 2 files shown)


karthikprasad:
> So I gave up and disabled memory benchmarking,

I don't know well enough to make suggestion to resolve the issue, but I'm not so sure about disable the check (atleast not without knowing what failed). 

>  since it seems like it's not a complicated thing to do due to some PyTorch update

Are you saying memory benchmarking is not necessary, or that doing it is easy?

karthikprasad (approved):

PR #684, opened 2024-11-04
delete CircleCI configs since GithubActions CI are now live

HuanyuZhang:
Summary: Cleanup job after #620

Differential Revision: D53202231


HuanyuZhang:
Close the PR since it will be solved by another one.

PR #695, opened 2024-12-09
Updates to Opacus to handle extra args returned in ghost clipping, dynamically provide epsilon_tolerance.

pagarwl:
## Types of changes

a) Currently make_private returns `criterion` if the ghost clipping algorithm was chosen, otherwise it doesn't. The return args list `Tuple[GradSampleModule, DPOptimizer, DataLoader]` is incorrect in this case. This causes compile time error when using ghost clipping technique. Removing the return arg specification (similar to make_private_with_epsilon), as a quick fix to avoid compile errors, while the approach can be further updated.

b) Allow providing epsilon_tolerance dynamically. 

Compile, unit tests, and manual runs.

## Checklist

### DIFF
--- a/opacus/grad_sample/embedding_norm_sample.py
+++ b/opacus/grad_sample/embedding_norm_sample.py
@@ -1,0 +1,150 @@
+#!/usr/bin/env python3
+# Copyright 2024, The Opacus authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Utility for computing gradient norm for the embedding layer.
+
+Based on the algorithm from the paper:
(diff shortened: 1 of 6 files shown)


opacus/tests/grad_samples/embedding_norm_sample_test.py
```diff
@@ -0,0 +1,218 @@
...
+        # Call the function we want to test
+        result = embedding_norm_sample.compute_embedding_norm_sample(
+            embedding_layer, activations, backprops
+        )
+
+        # Expected output based on the example
+        expected_norms = torch.tensor(
+            [0.0150, 0.0071, 0.0005, 0.0081, 0.0039], dtype=torch.float32
+        )
+        print("expected_norms: ", expected_norms)
```
EnayatUllah, line 214: Remove the print statement

opacus/grad_sample/embedding_norm_sample.py
```diff
@@ -0,0 +1,150 @@
...
+    """
+    device = activations[0].device
+    input_ids = activations[0].to(device)
+    grad_values = backprops.to(device)
+
+    # Reshape input_ids preserving the batch size as the first dimension
+    input_ids = input_ids.reshape(input_ids.shape[0], -1)
+
+    # Reshape grad_values preserving the embedding dimension as the last dimension
+    grad_values = grad_values.reshape(-1, grad_values.size(-1))
```
HuanyuZhang, line 113: @pagarwl  just double checking, this implementation not only works for 2 dimensions ([batch_size, seq_length]), but also works for any general dimension, correct?

PR #709, opened 2025-01-09
Enable multigpu_gradcheck unit test in CI

HuanyuZhang:
Summary: Not sure why the GPU unit tests have been disabled. Just use this diff to turn it on.

Differential Revision: D67957719


HuanyuZhang:
No CI GPU resources available for >1 gpu

PR #716, opened 2025-01-16
Modifying DPLossFastGradientClipping to add support for generative tasks with ghost clipping

EnayatUllah:
Summary:
1. Generative tasks for NLP output predictions of shape (B,T,C) i.e., (batch_size, sequence_length, vocab_size). To compute the cross-entropy loss in this case, usually the predictions are reshaped to (BxT, C) and targets to (BxT). This creates an issue with Ghost Clipping per sample loss computation as BxT is seen as the batch_size. In particular, the current implementation of Ghost Clipping results in loss_per_sample, coeff variables to have a shape of BxT and B respectively. This causes a shape mismatch error. This diff fixes that error by collapsing the loss_per_sample variable to shape B i.e., the loss across the sequence_length dim is averaged/summed.
2. ignore_index is also needed for the generative tasks as the task sometimes needs to ignore specific dummy tokens.

Differential Revision: D68047256

### DIFF
--- a/opacus/utils/fast_gradient_clipping_utils.py
+++ b/opacus/utils/fast_gradient_clipping_utils.py
@@ -68,7 +68,7 @@ class DPTensorFastGradientClipping:
         reduced_loss.backward(retain_graph=True)
         self.optimizer.zero_grad()
         coeff = self.module.get_clipping_coef()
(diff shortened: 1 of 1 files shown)


HuanyuZhang:
The diff has been committed.

PR #724, opened 2025-01-27
Add per-sample gradient norm computation as a functionality

iden-kalemaj:
Summary:
Per-sample gradient norm is computed for Ghost Clipping, but it can be useful generally. Exposed it as a functionality.

```
...

loss.backward()
per_sample_norms  = model.per_sample_gradient_norms

```

Differential Revision: D68634969


anhnami:
Great news!

PR #747, opened 2025-03-24
Diasble distributed per-layer clipping with hooks grad sample mode

iden-kalemaj:
Summary:
We disable support for distributed per-layer clipping with "hooks" grad sample mode, since it raises an error when using `register_full_backward_hook`. Distributed per-layer clipping with "ew" grad sample mode can still be used.

The issue arises because DistributedPerLayerOptimizer uses per-parameter hooks on top of the per-module hooks. During the backward pass, the per-parameter hooks fire before the per-module hooks. Per-sample gradients are computed when the per-module hooks fire, and an error occurs when the per-parameter hooks try to access the per-sample gradients before they are computed. Forcing the order in which hooks are called is not possible with PyTorch.

In the future, we may consider an approach that does not use per-parameter hooks. Given the limited use of per-layer clipping, we disable the faulty case of distributed data parallel with "hooks".

Differential Revision: D71706681

### DIFF
--- a/opacus/grad_sample/grad_sample_module.py
+++ b/opacus/grad_sample/grad_sample_module.py
@@ -207,7 +207,7 @@ class GradSampleModule(AbstractGradSampleModule):
             )
 
             self.autograd_grad_sample_hooks.append(
-                module.register_backward_hook(
+                module.register_full_backward_hook(
                     partial(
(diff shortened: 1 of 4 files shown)


HuanyuZhang:
launched a fix in a different PR, so abandon this one.

PR #826, opened 2026-06-01
Fix/pin peft version

### LINKED ISSUE #820: Silent corrupted LoRA weight updates with Opacus 1.5.4 + PEFT 0.18.x: training appears normal but models are unusable
## Environment
- opacus: 1.5.4
- peft: 0.18.1
- torch:  2.11.0+cu128 
- Python: 3.10

## What Happened

I ran DP-LoRA fine-tuning across 6 runs (2 models × 3 epsilon values, ~45 GPU-hours 
total). Training appeared completely normal throughout:
- Loss decreased as expected
- Epsilon accumulated correctly  
- No errors or warnings at any point
- Checkpoints saved successfully

All 6 models were unusable at inference time.

## Minimal Reproduction

```python
from peft import LoraConfig, get_peft_model
from transformers import AutoModelForCausalLM
from opacus import PrivacyEngine
import torch, torch.optim as optim
from torch.utils.data import DataLoader, TensorDataset

model = AutoModelForCausalLM.from_pretrained("gpt2", dtype=torch.float32)
config = LoraConfig(r=16, lora_alpha=32, lora_dropout=0.0,
    target_modules=["c_attn"], task_type="CAUSAL_LM")
model = get_peft_model(model, config)
optimizer = optim.AdamW(
    [p for p in model.parameters() if p.requires_grad], lr=1e-4)

dummy = torch.randint(0, 1000, (4, 32))
dl = DataLoader(TensorDataset(dummy), batch_size=4)

pe = PrivacyEngine(accountant='rdp')
model, optimizer, dl = pe.make_private_with_epsilon(
    module=model, optimizer=optimizer, data_loader=dl,
    target_epsilon=8.0, target_delta=1e-5, epochs=1, max_grad_norm=1.0,
)

param_name = [n for n in dict(model._module.named_parameters()) 
              if 'lora_B' in n][0]
before = dict(model._module.named_parameters())[param_name].clone()

model.train()
optimizer.zero_grad()
batch = dummy.cuda() if torch.cuda.is_available() else dummy
out = model(input_ids=batch, labels=batch)
out.loss.backward()
optimizer.step()

after = dict(model._module.named_parameters())[param_name]
print(f"Weight changed: {not torch.allclose(before.cpu(), after.cpu())}")
print(f"Max delta: {(after.cpu() - before).abs().max():.8f}")
```

**With PEFT 0.18.1:**
Weight changed: False
Max delta: 0.00000000

**With PEFT 0.13.2:**
Weight changed: True
Max delta: 0.00010004

The only way to detect the failure was:
1. Noticing identical utility collapse across all epsilon values 
   simultaneously (genuine epsilon-dependent collapse would scale with ε)
2. Manually comparing saved adapter weight norms against known 
   initialization values
3. Running inference and observing broken output
## Suspected Root Cause

PEFT 0.18.x changed LoRA parameter naming from `lora_A.weight` to 
`lora_A.default.weight` by introducing a `ModuleDict` with a named 
adapter key. Opacus 1.5.4 was written before this change and does not 
correctly handle the new parameter structure when registering per-sample 
gradient hooks via `GradSampleModule`.

The exact mechanism varies by architecture as some models show near-zero 
updates, others show partial corrupted updates ,but all produce 
unusable models.

## Workaround

```bash
pip install peft==0.13.2
```

Confirmed working across full multi-epoch training runs on both 2 models.

## Suggestion

Opacus's LoRA+PEFT tutorial should specify a compatible PEFT version, 
or add a version compatibility check that warns users when an 
incompatible PEFT version is detected.

MN-NR:
## Type of change
- Docs change / refactoring

Fixes #820

peft 0.14+ breaks Opacus hook registration when used with LoRA (see 
issue #820 for full root cause analysis and reproducer). The tutorial 
currently does not pin a peft version, causing users to silently 
install an incompatible version and hit the bug.

This commit pins peft==0.13.2 in the LoRA tutorial as suggested by 
@iden-kalemaj in issue #820, until full compatibility with peft 0.14+ 
is restored.

## Changes
- Added `!pip install peft==0.13.2` cell in 
  `tutorials/building_text_classifier.ipynb` before the LoRA section

# How Has This Been Tested
Verified peft==0.13.2 imports and LoRA setup works correctly locally:
- PyTorch 2.11.0, Python 3.10
- Full notebook not re-run (training takes too long)

### DIFF
--- a/opacus/privacy_engine.py
+++ b/opacus/privacy_engine.py
@@ -409,6 +409,18 @@ class PrivacyEngine:
         )
         if poisson_sampling:
             module.forbid_grad_accumulation()
+            
+        if hasattr(data_loader.dataset, '__len__'):
+            true_dataset_size = len(data_loader.dataset)
+            print(f"DEBUG: true_dataset_size = {true_dataset_size}")
+        else:
+            raise ValueError(
+                "Dataset must have __len__ for privacy accounting. "
+                "IterableDataset is not supported."
+    )
+        original_batch_size = data_loader.batch_size
+        if original_batch_size is None:
+            raise ValueError("DataLoader must have a batch_size")
 
         data_loader = self._prepare_data_loader(
             data_loader,
@@ -418,8 +430,8 @@ class PrivacyEngine:
             rand_on_empty=rand_on_empty,
         )
 
-        sample_rate = 1 / len(data_loader)
-        expected_batch_size = int(len(data_loader.dataset) * sample_rate)
+        sample_rate = original_batch_size / true_dataset_size
+        expected_batch_size = int(true_dataset_size * sample_rate)
 
         # expected_batch_size is the *per worker* batch size
         if distributed:
--- a/opacus/privacy_engine.py
+++ b/opacus/privacy_engine.py
@@ -409,10 +409,9 @@ class PrivacyEngine:
         )
         if poisson_sampling:
             module.forbid_grad_accumulation()
-            
+
         if hasattr(data_loader.dataset, '__len__'):
-            true_dataset_size = len(data_loader.dataset)
-            print(f"DEBUG: true_dataset_size = {true_dataset_size}")
+            true_dataset_size = len(data_loader.dataset)S
         else:
             raise ValueError(
                 "Dataset must have __len__ for privacy accounting. "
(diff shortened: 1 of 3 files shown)


iden-kalemaj:
@MN-NR before landing this, can you please check my comment in #820?

iden-kalemaj:
@MN-NR also this commit's changes seem to have been merged with another commit.

iden-kalemaj:
Closing this PR as the root cause is not the peft version, but rather the fact that `model.to(device)` should happen before `get_peft_model()`. See discussion in #820.

PR #825, opened 2026-05-29
Fix batch_first paths in DPMultiheadAttention

immu4989:
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Docs change / refactoring / dependency upgrade

## Motivation and Context / Related issue

Fixes #774 (plus a closely-related `batch_first=True` regression that surfaced while adding the requested test coverage).

`DPMultiheadAttention.forward()` had two `batch_first`-related issues:

**1. `attn_mask` shape validation ignored `batch_first` (#774).** The check used `query.size(0)` / `key.size(0)` regardless of `batch_first`. When `batch_first=True` those return the batch size, so any valid 2D or 3D mask was incorrectly rejected. Fix: compute `tgt_len` and a parallel `key_seq_len` in a `batch_first`-aware way and reference them in the mask validation.

**2. Output reshape under `batch_first=True` permuted nothing.** The reshape called `.view(bsz, tgt_len, embed_dim)` directly on the `(bsz * num_heads, tgt_len, head_dim)` tensor without first transposing the head and sequence dims, so head positions got spliced into time positions — empirically `max|diff|` against `nn.MultiheadAttention` ≈ 0.96. Fix: explicit `(B*H, L, D) → (B, H, L, D) → (B, L, H, D) → (B, L, E)` via `view → transpose(1, 2) → contiguous → view`. The `batch_first=False` branch is unchanged.

Neither bug surfaced previously because the existing `test_attn` / `test_dp_attn` only exercise `batch_first=False` and never pass `attn_mask`. See #774 for the diagnostic discussion — @iden-kalemaj greenlit bundling both fixes here.

## How Has This Been Tested

Added `test_attn_mask_batch_first` (Hypothesis-based, matching the existing `test_attn` / `test_dp_attn` style) covering both `batch_first` modes and both 2D and 3D mask shapes, asserting output and gradient parity with `nn.MultiheadAttention`.

Verified locally on Python 3.12 / pytest 9.0.3:

- `opacus/tests/dp_layers/dp_multihead_attention_test.py` — 4 tests pass (3 existing + new `test_attn_mask_batch_first`)
- `opacus/tests/grad_samples/dp_multihead_attention_test.py` — passes
- `opacus/tests/validators/multihead_attention_test.py` — both validator tests pass

## Checklist

- [x] The documentation is up-to-date with the changes I made.
[...]

PR #830, opened 2026-06-25
Add tutorial: DP fine-tuning of causal LMs with LoRA

### LINKED ISSUE #827: Add tutorial: DP fine-tuning of causal language models with LoRA
## Motivation

Opacus has one tutorial covering LoRA ([`tutorials/building_text_classifier.ipynb`](https://github.com/meta-pytorch/opacus/blob/main/tutorials/building_text_classifier.ipynb)), which uses BERT for sequence classification. There's no tutorial covering DP fine-tuning of causal language models (GPT-style) — which is the more common DP+PEFT use case for new users in 2026, and the surface that exposed the silent-corruption bug in #820.

`examples/fsdp_llama3.py` covers production-scale DP+LoRA on Llama with FSDP, but it's a runnable script, not a step-by-step notebook.

## Proposed scope

A new notebook `tutorials/building_dp_lora_for_llms.ipynb` that:

1. Uses a small causal LM (e.g., GPT-2) so it runs on a single-GPU laptop or CPU
2. Walks through LoRA setup → `make_private_with_epsilon` → training loop → sample generation
3. **Embeds the #820 safety patterns**:
   - PEFT version compat note (referencing PR #826 once merged)
   - Post-step weight-delta sanity check (the heuristic from #820)
   - `model.cuda()` placement note (matching the diagnostic direction from #820, pending CUDA-side confirmation)
4. Same tutorial format as the existing `tutorials/*.ipynb` notebooks

## Why this matters

- Causal-LM DP fine-tuning is what most users reach for when combining opacus + PEFT
- The #820 silent-corruption pattern is invisible to users without a "did the weights move?" check; embedding it in the canonical tutorial prevents repeat reports
- Builds on the safety work already in flight (PR #826)

## Question for maintainer

@iden-kalemaj — does this scope fit? Happy to wait for PR #826 to land before authoring (so the PEFT version pin pattern is established upstream), or to proceed with a self-contained `!pip install peft==0.13.2` cell. Let me know which you'd prefer.

immu4989:
## Types of changes

- [x] Docs change / refactoring / dependency upgrade

Closes #827.

Opacus has one LoRA tutorial today ([`tutorials/building_text_classifier.ipynb`](https://github.com/meta-pytorch/opacus/blob/main/tutorials/building_text_classifier.ipynb)), covering BERT for sequence classification. There's no tutorial for the more common contemporary case: causal language model fine-tuning with DP-SGD + LoRA. This PR adds that as `tutorials/building_dp_lora_for_llms.ipynb`.

The tutorial also bakes in the silent-corruption avoidance pattern surfaced and root-caused in #820. Three independent confirmations across CPU, Kaggle T4, and RTX 5090 settled it as a device-placement ordering issue (`model.to(device)` before `get_peft_model()`). The tutorial demonstrates the safe ordering explicitly with inline rationale and links back to #820.

## What's in the tutorial

- E2E NLG as the downstream task (small, standard in DP-NLP, used in the DiSK paper benchmark suite)
- GPT-2-small as the base model
- Two configurations compared: non-DP LoRA baseline vs DP LoRA at target epsilon=8
- BLEU, perplexity, throughput (tok/s), peak GPU memory, and final epsilon reported for each
- A dedicated "three safety patterns" section walking through:
  1. `model.to(device)` before `get_peft_model()` (#820 ordering)
  2. `model.train()` before `make_private_with_epsilon` (validator requirement)
  3. `poisson_sampling=False` (avoids GPT-2 forward-pass failure on empty Poisson batches)
- Honest discussion of when to reach for DP-LoRA, the learning-rate-needs-to-be-higher finding, and the engineering recipe to enable DP-full in a follow-up PR

## Canonical results on Kaggle T4 (embedded in the notebook)

| Config | BLEU | PPL | tok/s | Peak GB | Time | epsilon |
|---|---|---|---|---|---|---|
| non-DP LoRA | 24.72 | 1.60 | 4679 | 2.17 | 438 s | no DP |
| DP LoRA | 18.24 | 2.71 | 4599 | 2.65 | 445 s | 7.08 |

DP costs about 26% relative BLEU at epsilon ~= 7 on this task, with about 22% memory overhead and negligible throughput impact.

## What's intentionally out of scope

DP-full fine-tuning of GPT-2 with all 125M parameters trainable. Attempted via `opacus.validators.ModuleValidator.fix()` (Conv1D -> Linear swap) plus `grad_sample_mode='functorch'`; neither resolved a per-sample-gradient shape mismatch in `clip_and_accumulate` that appears to stem from GPT-2's tied `wte` / `lm_head` weights. The notebook's section 16 documents a recommended engineering recipe for a follow-up PR (untie weights, then re-apply fix and functorch). Felt cleaner to ship the LoRA portion as scoped here than to block on the DP-full work.

## How Has This Been Tested

End-to-end run on Kaggle T4 with pinned versions:

- `opacus 1.6.0`
- `peft 0.18.1`
- `transformers 5.0.0`
- `torch 2.10.0+cu128`

Cell outputs in the committed notebook reflect the actual run. The three safety patterns each addressed a real failure mode encountered during development.

## Checklist

- [x] I have read the **CONTRIBUTING** document and completed the CLA (CLA on file from PR #825).


tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,642 @@
...
+     },
+     "metadata": {}
+    }
+   ],
+   "execution_count": 10
+  },
+  {
+   "id": "74cf7106-b186-49c5-97e1-616ffa2d38a7",
+   "cell_type": "markdown",
+   "source": "## 11. Configurations to compare\n\nTwo configurations, both LoRA-based, identical in every respect except whether the `PrivacyEngine` is attached. The `lr=5e-4` for the DP variant is intentionally higher than the non-DP `1e-4`: DP-SGD's per-step update is dampened by gradient clipping and Gaussian noise, so a higher learning rate is needed to make comparable progress. Empirically, `lr=1e-4` produced BLEU near zero for the DP variant in our early runs.\n\nDP-full (no LoRA) is intentionally left out of this tutorial; see the follow-up section for what is needed to enable it.",
```
iden-kalemaj: DP-SGD needing a higher learning rate seems inconsistent with prior work which observes that DP-SGD usually needs a smaller learning rate to account for the gradient noise. If loss is not converging, two hyper-parameters to test would be: increasing batch size compared to the non-private training and tuning max_grad_norm to a higher value.

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,642 @@
...
+    "\n",
+    "Memory overhead from opacus is modest at LoRA's parameter scale. About 22% more peak GPU memory (2.17 GB to 2.65 GB) covers the per-sample-gradient accumulation. This stays comfortably within a T4's 16 GB envelope and would not constrain a typical fine-tuning workflow.\n",
+    "\n",
+    "Throughput is essentially unchanged. Both configurations hit roughly 4600 tokens per second; opacus is not a throughput bottleneck for LoRA at this size.\n",
+    "\n",
+    "**Learning rate matters more for DP.** The 5\u00d7 higher lr for the DP variant is not optional. Leave both configurations at `lr=1e-4` and the DP variant produces BLEU near zero despite reaching reasonable perplexity. The intuition: gradient clipping and Gaussian noise both shrink the effective per-step update, so the optimizer needs more aggressive learning to make comparable progress per step. This is consistent with the DP-NLP literature and worth surfacing explicitly in any production setup."
```
iden-kalemaj: do you have any papers to point to here about the higher learning rate used by DPS-GD?

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,642 @@
...
...orch/opacus/issues/827)\n\n## What you'll get out of this notebook\n\n1. A working recipe for combining `opacus`, `peft`, and HuggingFace `transformers` on a causal LM\n2. The three device/training-mode ordering patterns that prevent silent corruption (see [#820](https://github.com/meta-pytorch/opacus/issues/820))\n3. A concrete sense of the utility cost of DP at a fixed privacy budget on this task\n4. Honest notes on what does and does not work yet for full-finetune DP on GPT-2 (out of scope for this tutorial, recipe documented at the end)\n\n## Prerequisites\n\n- Familiarity with the standard DP-SGD setup (`PrivacyEngine`, `make_private_with_epsilon`). The opacus [Building text classifier tutorial](https://github.com/meta-pytorch/opacus/blob/main/tutorials/building_text_classifier.ipynb) is a good warm-up.\n- Some prior exposure to LoRA. The [original paper](https://arxiv.org/abs/2106.09685) is short and worth reading.\n\n## Environment and runtime\n\n- Single GPU is sufficient (Kaggle T4 used here)\n- Roughly 15 minutes end-to-end on T4 at the settings below (most of it is the 2000-step training pass)\n- Pinned versions: `opacus>=1.6.0`, `peft>=0.18,<0.19`, `transformers>=5.0`",
```
iden-kalemaj: Let's please edit language throughout the tutorial that sounds LLM-like such as "honest notes" or "the price of admission"

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,642 @@
...
...nto the *learned* parameters. LoRA is a natural fit: it constrains updates to a small low-rank subspace alongside the frozen base weights. In our experiments below, LoRA trains roughly **0.47%** of the GPT-2-small parameters and still reaches a respectable BLEU on the E2E NLG benchmark, both with and without DP.\n\nThe combination also lines up cleanly with how practitioners deploy DP today. Sensitive training data warrants a real privacy guarantee; production training budgets warrant parameter-efficient methods. DP-LoRA is the intersection.\n\n## The task: E2E NLG\n\n[E2E NLG](https://arxiv.org/abs/1706.09254) is a small structured-data-to-text generation task originally from the 2017 E2E challenge. Inputs are slot-value meaning representations such as `name[The Vaults], eatType[pub], priceRange[more than \u00a330]`; outputs are short natural-language descriptions. The dataset is small enough to fine-tune in minutes on a single GPU but rich enough that BLEU separates a trained model from chance.\n\nIt is also the standard benchmark used in the [DiSK paper](https://arxiv.org/abs/2410.03883) and adjacent DP-NLP work, which makes results here comparable to the published literature.",
```
iden-kalemaj: in the DISK paper the reported BLEU score is >60%, the values of 20-30% are kind of low in the tutorial. would it be possible to train for longer?

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,642 @@
...
...        # \u2190 move base model to CUDA FIRST\nmodel = get_peft_model(model, config)   # then apply LoRA (LoRA params land on CUDA too)\n# then PrivacyEngine.make_private(...)\n```\n\n### Pattern B: `model.train()` before `make_private_with_epsilon()`\n\nopacus's `ModuleValidator` (1.6.0+) raises `IllegalModuleConfigurationError(\"Model needs to be in training mode\")` if the model is not in train mode when `make_private_with_epsilon` runs. `get_peft_model()` puts the model in eval mode by default, so an explicit `model.train()` call between PEFT wrapping and opacus wrapping is required.\n\n### Pattern C: `poisson_sampling=False`\n\nopacus's default Poisson sampling occasionally produces an empty batch (probability around `e^(-batch_size)` per step). GPT-2's forward pass calls `attention_mask.view(batch_size, -1)`, which fails for a zero-element tensor because the `-1` dimension is ambiguous. Setting `poisson_sampling=False` switches to uniform-without-replacement, a valid DP-SGD variant (this is what the original [Abadi 2016](https://arxiv.org/abs/1607.00133) paper uses), with deterministic batch sizes and no empty-batch edge case. The accountant handles both regimes correctly.",
```
iden-kalemaj: "a valid DP-SGD variant" is misleading here. The privacy analysis of DP-SGD assumes poisson sampling but in practice poisson sampling is often disabled.  See the warning produced when possion sampling is disabled.

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,642 @@
...
+     "shell.execute_reply": "2026-06-25T05:04:15.867984Z"
+    }
+   },
+   "outputs": [],
+   "execution_count": 8
+  },
+  {
+   "id": "2851ac76-8873-4d92-a60f-2df2d44bb26c",
+   "cell_type": "markdown",
+   "source": "## 9. The shared training function\n\nSame loop for all configurations. The `cfg.dp` block adds the `PrivacyEngine`, applies the three safety patterns, and uses `poisson_sampling=False` per Pattern C. Memory and throughput are tracked across the run; the final epsilon is read from the accountant at the end. The per-step skip on empty batches is a defensive guard that should never actually fire with `poisson_sampling=False`.",
```
iden-kalemaj: lets remove the per-step empty skip since its not adding value here.

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,642 @@
...
...alse,\n            # but harmless guard)\n            if batch['input_ids'].numel() == 0:\n                continue\n            batch = {k: v.to(device) for k, v in batch.items()}\n            outputs = model(**batch)\n            loss = outputs.loss\n            optimizer.zero_grad()\n            loss.backward()\n            optimizer.step()\n            loss_history.append(loss.item())\n            tokens_processed += batch['input_ids'].numel()\n            if step % 100 == 0:\n                print(f'  step {step:4d}  loss={loss.item():.4f}')\n            step += 1\n\n    t_total = time.perf_counter() - t_start\n    peak_mem_gb = (torch.cuda.max_memory_allocated() / 1024**3) if torch.cuda.is_available() else 0.0\n    final_epsilon = privacy_engine.get_epsilon(delta=cfg.target_delta) if cfg.dp else float('inf')\n\n    return {\n        'config': cfg,\n        'model': model,\n        'mean_loss': sum(loss_history) / len(loss_history),\n        'final_loss': loss_history[-1],\n        'tokens_per_sec': tokens_processed / t_total,\n        'peak_mem_gb': peak_mem_gb,\n        'wall_clock_sec': t_total,\n        'epsilon': final_epsilon,\n        'steps_completed': step,\n    }\n",
```
iden-kalemaj: " Uniform sampling is a documented opacus mode and a standard DP-SGD variant" -> "Uniform sampling with DP-SGD is commonly used in practice".

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,642 @@
...
+    "Sample run on Kaggle T4 (your numbers will vary slightly with random seeds, hardware, and library minor versions):\n",
+    "\n",
+    "| Config | BLEU | PPL | tok/s | Peak GB | Time | \u03b5 |\n",
+    "|---|---|---|---|---|---|---|\n",
+    "| non-DP LoRA | 24.72 | 1.60 | 4679 | 2.17 | 438 s | no DP |\n",
+    "| DP LoRA | 18.24 | 2.71 | 4599 | 2.65 | 445 s | 7.08 |\n",
+    "\n",
+    "The non-DP LoRA baseline scores BLEU 24.72 on E2E NLG after 2000 steps while training only 589K of GPT-2-small's 125M parameters (about 0.47%). That is the parameter-efficiency story that motivates LoRA in the first place: most of the useful signal can be captured in a small low-rank perturbation, and the rest of the network does not need to move.\n",
+    "\n",
+    "DP costs roughly 26% relative BLEU at \u03b5 \u2248 7 here, going from 24.72 down to 18.24. Perplexity moves from 1.60 to 2.71. The model still clearly learns the conditional generation task; the noise from DP-SGD shifts utility downward but does not collapse it. The 26% figure is in the ballpark of what comparable papers report at similar privacy budgets on this task.\n",
```
iden-kalemaj: "The 26% figure is in the ballpark of what comparable papers report at similar privacy budgets on this task" -> is there a reference for this?

tutorials/building_dp_lora_for_llms.ipynb
iden-kalemaj: "It is also the standard benchmark used in the [DiSK paper]" -> "It is also one of the benchmarks used in the [DiSK paper]"

tutorials/building_dp_lora_for_llms.ipynb
immu4989: you are right that the classic guidance points the other way for training from scratch, where noise dominates. For DP fine tuning of LMs specifically, though, the reported optima tend to be larger than non private: Li et al. (ICLR 2022, arXiv:2110.05679) and Yu et al. (ICLR 2022, arXiv:2110.06500, which is GPT-2 + LoRA on E2E, essentially this tutorial's setting) both fine tune with learning rates roughly an order of magnitude above typical non private values. The mechanism in our run is clipping rather than noise: with max_grad_norm=1.0 early per sample grads are heavily clipped, so the average update is attenuated well below its non private magnitude, and lr compensates for that attenuation.

That said, your two knobs attack the same problem at the source, so I will run the ablation in the revision: fixed lr=1e-4 with (a) larger batch and (b) max_grad_norm raised, alongside the current lr=5e-4 setting, and present whichever configuration tells the cleanest story. I will also reword the current framing, which incorrectly lumps noise in with clipping as the reason for the higher lr.

tutorials/building_dp_lora_for_llms.ipynb
immu4989: Yes: Li et al., "Large Language Models Can Be Strong Differentially Private Learners" (ICLR 2022, arXiv:2110.05679), which studies lr/batch interplay for DP LM fine-tuning directly, and Yu et al., "Differentially Private Fine-tuning of Language Models" (ICLR 2022, arXiv:2110.06500) for the GPT-2 + LoRA + E2E setting specifically. I'll cite both at this point in the text and soften the sentence to describe the clipping attenuation mechanism rather than assert a general rule.

tutorials/building_dp_lora_for_llms.ipynb
immu4989: Will do a full editing pass and remove that phrasing throughout.

tutorials/building_dp_lora_for_llms.ipynb
immu4989: The gap has two causes and I will fix both. First, training budget: the current notebook runs 2000 steps, about 0.4 epochs; the papers train for multiple epochs. Second, and larger: evaluation protocol. The official E2E metric is multi reference BLEU (each MR has several human references) with beam decoding, while the notebook currently computes single reference sacreBLEU with greedy decoding. On E2E that protocol difference alone is worth a large number of BLEU points, so the current 24.72 isn't comparable to the >60 figures in DiSK or Yu et al.

Revision plan: train several epochs (fits fine on the T4 budget), group the references per MR for multi reference BLEU, decode with beam search, and report the new table. If the numbers still land meaningfully below the published range for GPT-2-small at comparable ε I will say so explicitly rather than paper over it.

tutorials/building_dp_lora_for_llms.ipynb
immu4989: Agreed, that wording overclaims. The accountant's amplification analysis assumes Poisson subsampling, and with uniform batches the reported ε rests on an approximation that the sampling scheme doesn't strictly satisfy (Chua et al., ICML 2024, arXiv:2403.17673 quantify the gap). I will adopt your phrasing, reproduce the opacus warning verbatim in the notebook output so readers see it, and add a sentence stating that reported ε under uniform sampling should be treated as approximate. Happy to also mention the alternative of keeping Poisson sampling and handling the empty batch edge case at the loop level if you would prefer the tutorial teach that path instead.

tutorials/building_dp_lora_for_llms.ipynb
immu4989: Will remove in the next revision

tutorials/building_dp_lora_for_llms.ipynb
immu4989: ok. will use the wordings accordingly

tutorials/building_dp_lora_for_llms.ipynb
immu4989: No, and on reflection the claim doesn't survive the protocol fix from your other comment: with proper multi reference evaluation and adequate training, Yu et al. report a much smaller relative gap at comparable ε than 26%. I'll delete the sentence and instead compare our revised numbers directly against their GPT-2/E2E table, whatever that comparison ends up showing.

tutorials/building_dp_lora_for_llms.ipynb
immu4989: Fixed as suggested.

immu4989:
Thanks for the careful review. Summary of the revision, in rough order of effort:

1. Language pass removing the LLM sounding phrasing throughout.
2. Longer training (multiple epochs) plus a corrected evaluation protocol: multi-reference BLEU with beam decoding, matching the official E2E metric, so numbers are comparable to DiSK and Yu et al.
3. Learning rate section rewritten around the clipping attenuation mechanism with citations (Li et al. 2022, Yu et al. 2022), plus the batch-size / max_grad_norm ablation you suggested at fixed lr.
4. Poisson sampling discussion corrected per your wording, with the accountant caveat made explicit and the empty batch skip removed.
5. Unsupported comparison claim replaced with a direct comparison against published GPT-2/E2E numbers.

Reruns are on Kaggle T4 so expect the revised notebook within a few days.

immu4989:
Revision pushed. Every item from the review, plus the numbers.

**Evaluation protocol.** The notebook now scores multi-reference BLEU with beam-5 decoding over 100 held-out MRs, matching the official E2E protocol, and trains two full epochs over the whole train split instead of 2000 steps. Together those took the non-private run from 24.72 to **68.38** on the same model and code. That answers your question about the DiSK numbers: the gap was almost entirely the scoring protocol, not the training.

**Results on a T4:**

| Config | BLEU | PPL | Peak GB | ε |
|---|---|---|---|---|
| non-DP LoRA (lr 1e-4) | 68.38 | 1.50 | 2.17 | no DP |
| DP LoRA (lr 5e-4) | 47.13 | 2.02 | 2.65 | 8.00 |
| DP LoRA (lr 1e-4, batch 32) | 44.88 | 2.81 | 8.16 | 7.99 |
| DP LoRA (lr 1e-4, clip 3.0) | 38.90 | 3.08 | 3.66 | 8.00 |

**On the learning rate.** Your suggested knobs held up well, and better than my original framing did. Raising the batch size to 32 while keeping the non-private learning rate gets 44.88, close to the 47.13 from raising the learning rate fivefold, and both beat raising the clipping threshold at 38.90. The tutorial now presents all three as viable routes with that ranking rather than recommending the learning rate specifically, and notes that batch size and learning rate work through different mechanisms so a joint sweep would likely beat either alone. The section is rewritten around clipping attenuation with citations to Li et al. (2110.05679) and Yu et al. (2110.06500).

**Poisson sampling.** Adopted your wording, added the caveat that the accountant's analysis assumes Poisson sampling so the reported ε under uniform sampling is an approximation, with a pointer to Chua et al. (2403.17673). The per-step empty-batch skip is gone.

**The unsupported comparison claim is deleted.** I could not verify E2E-specific BLEU tables for either DiSK or Yu et al. from the sources I could reach — the numbers in the Yu et al. abstract are DART, not E2E — so rather than substitute another unverified figure, the tutorial now states what the results can and cannot be compared against and leaves the published-comparison anchor at the level you cited.

Also did a language pass throughout, and batch 64 turned out to exceed 16 GB on a T4 once per-sample gradients are counted, so the batch ablation uses 32 and the memory cost is noted.

immu4989:
The three GitHub Actions workflows expired after waiting 30 days for fork-workflow approval; they contain zero jobs, so these are not test failures. I’ve retriggered them. When convenient, could you approve the new workflow runs so CI can execute?

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,6747 @@
...
+     "status": "completed"
+    },
+    "tags": []
+   },
+   "source": [
+    "# Differentially Private LoRA Fine-Tuning of Causal Language Models\n",
+    "\n",
+    "Walk through how to fine-tune a GPT-style causal language model on a small generation task with differential privacy, using parameter-efficient LoRA adapters. We compare a non-private LoRA baseline against the DP-LoRA variant and report BLEU, perplexity, peak memory, and throughput for each.\n",
+    "\n",
+    "Tracking: [meta-pytorch/opacus#827](https://github.com/meta-pytorch/opacus/issues/827)\n",
```
iden-kalemaj: please remove "Tracking: [meta-pytorch/opacus#827](https://github.com/meta-pytorch/opacus/issues/827)\n",

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,6747 @@
...
+     "status": "completed"
+    },
+    "tags": []
+   },
+   "source": [
+    "## Why combine DP-SGD with LoRA?\n",
+    "\n",
+    "DP-SGD adds calibrated Gaussian noise to the per-sample gradient sum at each step. The noise scale grows with the L2 sensitivity of the per-sample gradient (the clipping threshold), so a model with fewer trainable parameters effectively reduces the noise injected into the *learned* parameters. LoRA is a natural fit: it constrains updates to a small low-rank subspace alongside the frozen base weights. In our experiments below, LoRA trains roughly **0.47%** of the GPT-2-small parameters and still reaches a respectable BLEU on the E2E NLG benchmark, both with and without DP.\n",
+    "\n",
+    "The combination also lines up cleanly with how practitioners deploy DP today. Sensitive training data warrants a real privacy guarantee; production training budgets warrant parameter-efficient methods. DP-LoRA is the intersection.\n",
```
iden-kalemaj: please remove "The combination also lines up cleanly with how practitioners deploy DP today. Sensitive training data warrants a real privacy guarantee; production training budgets warrant parameter-efficient methods. DP-LoRA is the intersection"

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,6747 @@
...
+    "\n",
+    "### Pattern C: `poisson_sampling=False`\n",
+    "\n",
+    "opacus's default Poisson sampling occasionally produces an empty batch (probability around `e^(-batch_size)` per step). GPT-2's forward pass calls `attention_mask.view(batch_size, -1)`, which fails for a zero-element tensor because the `-1` dimension is ambiguous. Setting `poisson_sampling=False` switches to uniform-without-replacement sampling, which has deterministic batch sizes and avoids the edge case. Uniform sampling with DP-SGD is commonly used in practice.\n",
+    "\n",
+    "One caveat to be aware of: the accountant's privacy analysis assumes Poisson sampling. opacus emits a warning when Poisson sampling is disabled, and the reported \u03b5 under uniform sampling should be treated as an approximation rather than an exact guarantee (Chua et al., ICML 2024, [arXiv:2403.17673](https://arxiv.org/abs/2403.17673) quantify the gap between the two sampling schemes). Keep the warning visible in your own runs instead of suppressing it.\n"
```
iden-kalemaj: please remove "Keep the warning visible in your own runs instead of suppressing it."

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,6747 @@
...
+     "exception": false,
+     "start_time": "2026-07-20T11:48:18.736680+00:00",
+     "status": "completed"
+    },
+    "tags": []
+   },
+   "source": [
+    "## 2. Sanity-check the environment\n",
+    "\n",
+    "If CUDA reports unavailable, enable a GPU accelerator before proceeding. The assertions guard against pip silently resolving to incompatible versions."
```
iden-kalemaj: please remove "The assertions guard against pip silently resolving to incompatible versions" as there are no assertions.

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,6747 @@
...
+    "| non-DP LoRA (lr 1e-4) | 68.38 | 1.50 | 4776 | 2.17 | 2254 s | no DP |\n",
+    "| DP LoRA (lr 5e-4) | 47.13 | 2.02 | 4704 | 2.65 | 2289 s | 8.00 |\n",
+    "| DP LoRA (lr 1e-4, batch 32) | 44.88 | 2.81 | 4652 | 8.16 | 2315 s | 7.99 |\n",
+    "| DP LoRA (lr 1e-4, clip 3.0) | 38.90 | 3.08 | 4705 | 3.66 | 2288 s | 8.00 |\n",
+    "\n",
+    "LoRA trains 589,824 parameters, about 0.47% of GPT-2-small's 125M, and the non-private run reaches BLEU 68.38. That is the parameter-efficiency argument in a single number.\n",
+    "\n",
+    "Differential privacy at \u03b5 = 8 costs roughly 31% relative BLEU here, 68.38 down to 47.13 for the best DP configuration, with perplexity moving from 1.50 to 2.02. The model still learns the conditional generation task; the noise and clipping shift utility down without collapsing it.\n",
+    "\n",
+    "### Three ways to pay for clipping, and they are not equal\n",
```
iden-kalemaj: please rephrase "Three ways to pay for clipping, and they are not equal". This sounds very akin to claude generated text, which makes users less likely to engage. Suggest a title like "Three ways to compensate from utility loss due to clipping".

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,6747 @@
...
+    "\n",
+    "- raising the learning rate fivefold, to 5e-4, recovers the most utility (47.13)\n",
+    "- raising the batch size fourfold, to 32, recovers nearly as much (44.88) while holding the learning rate at the non-private 1e-4\n",
+    "- raising the clipping threshold to 3.0 helps least (38.90)\n",
+    "\n",
+    "The first two are close enough that either is a reasonable starting point, and a sweep over both together would likely beat either alone. A larger batch improves the signal-to-noise ratio of the noisy gradient mean, which is a different mechanism from simply taking bigger steps, so the two are complementary rather than redundant.\n",
+    "\n",
+    "It is worth seeing how badly the uncompensated setting fails. At `lr=1e-4` with `max_grad_norm=1.0`, the same configuration produced BLEU near zero while still reaching a plausible-looking perplexity. Perplexity alone will not tell you the conditional task has failed to train.\n",
+    "\n",
+    "### Memory is where the batch-size route gets expensive\n",
```
iden-kalemaj: similar comment, rephrase to "Increasing batch size affects memory usage".

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,6747 @@
...
+    "\n",
+    "Peak memory for the batch-32 run is 8.16 GB against 2.65 GB at batch 8, and a batch of 64 exceeded the 16 GB available on a T4 once opacus's per-sample gradients were accounted for. If you take the larger-batch route to compensate for clipping, budget for it.\n",
```
iden-kalemaj: Please remove "If you take the larger-batch route to compensate for clipping, budget for it".

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,6747 @@
...
+    "\n",
+    "One caveat on those memory figures: they are recorded per configuration inside a single process, and the two batch-8 DP runs report 2.65 GB and 3.66 GB despite being identical in batch size. The clipping threshold does not affect allocation, so the difference reflects allocator state carried over from the preceding run rather than a real cost. Treat the numbers as indicative, and measure in isolation if you need them to be precise.\n",
+    "\n",
+    "### On comparing against published results\n",
+    "\n",
+    "Published E2E numbers generally come from larger GPT-2 variants, full fine-tuning rather than LoRA, and longer schedules, so the figures above are not directly comparable to them. The useful signal is that the non-private 68.38 sits in the range typically reported for well-trained models on this benchmark, which says the evaluation protocol here is sound. An earlier version of these experiments scored the same trained model at 24.72 using single-reference BLEU with greedy decoding; the entire difference came from the scoring protocol, not from the model.\n"
```
iden-kalemaj: please remove "An earlier version of these experiments scored the same trained model at 24.72 using single-reference BLEU with greedy decoding; the entire difference came from the scoring protocol, not from the model". Prior errors are not informative.

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,6747 @@
...
+    "\n",
+    "It is worth seeing how badly the uncompensated setting fails. At `lr=1e-4` with `max_grad_norm=1.0`, the same configuration produced BLEU near zero while still reaching a plausible-looking perplexity. Perplexity alone will not tell you the conditional task has failed to train.\n",
+    "\n",
+    "### Memory is where the batch-size route gets expensive\n",
+    "\n",
+    "Peak memory for the batch-32 run is 8.16 GB against 2.65 GB at batch 8, and a batch of 64 exceeded the 16 GB available on a T4 once opacus's per-sample gradients were accounted for. If you take the larger-batch route to compensate for clipping, budget for it.\n",
+    "\n",
+    "One caveat on those memory figures: they are recorded per configuration inside a single process, and the two batch-8 DP runs report 2.65 GB and 3.66 GB despite being identical in batch size. The clipping threshold does not affect allocation, so the difference reflects allocator state carried over from the preceding run rather than a real cost. Treat the numbers as indicative, and measure in isolation if you need them to be precise.\n",
```
iden-kalemaj: Please rephrase "One caveat on those memory figures" to "A caveat on reported memory utilization:"

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,6747 @@
...
+    "\n",
+    "```python\n",
+    "B, T = input_ids.shape\n",
+    "position_ids = torch.arange(T).unsqueeze(0).expand(B, T)\n",
+    "outputs = model(input_ids=input_ids, position_ids=position_ids, labels=input_ids)\n",
+    "```\n",
+    "\n",
+    "With that change, DP full fine-tuning of GPT-2 trains normally. The same pattern should apply to other models with learned absolute position embeddings that broadcast their position ids. A library-level fix in opacus's embedding grad sampler is under discussion.\n",
+    "\n",
+    "### Other polishes worth doing\n",
```
iden-kalemaj: please remove the section "    "### Other polishes worth doing\n","

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,6747 @@
...
+     "exception": false,
+     "start_time": "2026-07-20T14:24:44.493709+00:00",
+     "status": "completed"
+    },
+    "tags": []
+   },
+   "source": [
+    "## 16. Follow-up work\n",
+    "\n",
+    "### Enabling DP-full fine-tuning on GPT-2\n",
```
iden-kalemaj: please remove the section on ""### Enabling DP-full fine-tuning on GPT-2\n", this can be handled in another issue.

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,6747 @@
...
+    "tags": []
+   },
+   "source": [
+    "## 15. When to reach for DP-LoRA\n",
+    "\n",
+    "A practical heuristic, given the numbers above:\n",
+    "\n",
+    "Use DP-LoRA when the training data is sensitive enough to warrant a real privacy guarantee and the downstream task tolerates a meaningful utility drop relative to a non-DP baseline. The utility cost of DP at this budget is quantified in the results table above; for published reference points on this task, see Yu et al. (ICLR 2022).\n",
+    "\n",
+    "Consider full DP fine-tuning instead of DP-LoRA when LoRA's restricted parameter subspace is the bottleneck (not the privacy budget) and you have compute headroom to handle opacus's per-sample-gradient memory overhead at the full model scale. The full-finetune path on GPT-2 specifically also needs the engineering steps in the follow-up section below."
```
iden-kalemaj: please remove "The full-finetune path on GPT-2 specifically also needs the engineering steps in the follow-up section below"

tutorials/building_dp_lora_for_llms.ipynb
```diff
@@ -0,0 +1,6747 @@
...
+     "duration": 0.012894,
+     "end_time": "2026-07-20T14:24:44.479777+00:00",
+     "exception": false,
+     "start_time": "2026-07-20T14:24:44.466883+00:00",
+     "status": "completed"
+    },
+    "tags": []
+   },
+   "source": [
+    "## 15. When to reach for DP-LoRA\n",
```
iden-kalemaj: Let's remove the discussion "When to reach for DP-LoRA".

iden-kalemaj:
Hi @immu4989, apologies for the delayed review. I have done another pass on the language used in the tutorial and left some comments. For the tutorial to be engaging to readers, I think it is important to remove agent-generated-sounding language. Please see the comments and also edit where you see fit to remove patterns of speech that are common to LLMs.

PR #844, opened 2026-08-18
Patch neg eps bug and add test for each accountant

eddiestudies:
## Types of changes

Minimal clamping change to avoid returning negative epsilon values(rdp, prv) or failing(gdp)

Issue was raised [585](https://github.com/meta-pytorch/opacus/issues/585) for prv. 
Adds a clamp to avoid negative epsilons for rdp and prv. rdp had a similar negative epsilon issue. 
gdp failed because brentq did not find a sign change between [0, 500].

I added a test that previously failed for the three accountants.

## Checklist

### DIFF
--- a/opacus/accountants/analysis/gdp.py
+++ b/opacus/accountants/analysis/gdp.py
@@ -94,6 +94,9 @@ def eps_from_mu(*, mu: float, delta: float) -> float:
         """Reversely solve dual by matching delta."""
         return delta_eps_mu(eps=x, mu=mu) - delta
 
+    if f(0) <= 0:
+        return 0.0
+
     return optimize.root_scalar(f, bracket=[0, 500], method="brentq").root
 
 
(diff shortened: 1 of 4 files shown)


iden-kalemaj:
Hi @eddiestudies, I am a bit reluctant to land this change. The negative value indicates that the chosen parameters are probably not very meaningful. Maybe instead we can instead raise a warning that the negative epsilon value could be due to a very high value of delta, and offer to the user to consider lowering the value of delta for meaningful privacy guarantees. See also @HuanyuZhang's comment in the original issue.

eddiestudies:
I understand. A warning would definitely help.

 I tried to better understand the options and realized that summing epsilons might lead to unintended under reporting if some are negative.  I found other libraries clamp to 0 to avoid this but I do agree there is a need for a warning here. 

I can verify that internal summing within Opacus does not lead to bugs when eps is previously negative.

